Issue #11, opened 2017-09-26
Incorrect padding on logo


xPaw:
![26-155201704](https://user-images.githubusercontent.com/613331/30870194-e6116cd8-a2eb-11e7-8d2e-27a6aa9c2641.png)
1. Incorrect background on `:active`
2. There is no `:hover` style compared to other menu elements.

On mobile view, the logo is not vertically aligned:
![26-155318287](https://user-images.githubusercontent.com/613331/30870228-04456c68-a2ec-11e7-9d57-a301cb036694.png)

pcvonz:
I just pushed an update that should address those two issues. Can you confirm that they're fixed? Also, what browser are you using?

xPaw:
Chrome 61 on Windows 10.

The padding is still off (0 on the left side), and there is still no hover effect.
![26-164869046](https://user-images.githubusercontent.com/613331/30872736-c6752998-a2f3-11e7-89d1-fcbc44c05590.png)

akien-mga:
Partly fixed by #12, but now that I've updated the logo image I still see some missing left-padding:
![image](https://user-images.githubusercontent.com/4701338/32042242-289c40e8-ba37-11e7-8856-b770d95109d5.png)
cc @death180

akien-mga:
Ah nevermind, #12 is not merged yet. @pcvonz can you check it?

djrm:
closed by f69d1ca167d92252d00c3b3c967cb9a64dfdfdf1

PR #12, opened 2017-10-15
Update main.scss

ajack764:
Making the padding around the logo symmetrical as suggested in #11


pcvonz:
`margin-right: 0 .5rem` is an incorrect value for that property. Did you mean `margin: 0 .5rem`?

ajack764:
Wow I don't know how I made that mistake, but yeah it should be that

akien-mga:
Closing as it stalled.

PR #15, opened 2017-11-22
Make the main fonts bolder and smaller

Calinou:
This makes them much more readable, especially on Windows. This mostly affects news posts as well as the Download page.

Some previews on Firefox Nightly 59 on Windows 10 (with [MacType](https://github.com/snowie2000/mactype) installed and enabled):

readable_fonts_home_top
readable_fonts_home_bottom
readable_fonts_news_post
readable_fonts_download

### DIFF
--- a/themes/godotengine/src/scss/main.scss
+++ b/themes/godotengine/src/scss/main.scss
@@ -75,7 +75,7 @@ a.slink:hover {
   .jumbotron .button {
     display: block;
   }
-  
+
   .jumbotron h1 {
     font-size: 33px;
     line-height: 45px;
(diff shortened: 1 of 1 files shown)


Calinou:
It looks like this pull request doesn't affect the deployed website (perhaps the CSS isn't automatically rebuilt?). Can someone look into this? I don't know about the build process, but it seems hackish to commit generated files to a Git repository.

pcvonz:
@Calinou Yeah, it'd be awesome if we could set up a script to rebuild the SCSS when a change is pulled. What I've been doing is just SCPing my changes over to the live server.

PR #32, opened 2018-01-30
Added nice opacity hover effect to header, button and tab links

levrik:
image

image

image


djrm:
i would use the accent color instead.

djrm:
also some elements are already using the accent color, like the pagination.

levrik:
Was fixed by 512ebd976dd980ffee48a847f65b9c8135d2b752

PR #37, opened 2018-02-07
Grammatical fix

jcs224:
- The node-based architecture _belongs_ to Godot, so Godot needs to be a possessive.
- Put a dash between "node" and "based". This is a compound adjective.


akien-mga:
Thanks!

PR #38, opened 2018-02-07
Adverb + plural possessive

jcs224:
- "Truly" instead of "True." Truly is an adverb, which emphasizes how open Godot's development is.
- "others'" instead of "other's". An apostrophe between the word and letter "s" indicates a singular possessive, whereas putting the apostrophe _after_ the "s" indicates a plural possessive. This is the correct usage.


akien-mga:
Thanks!

PR #49, opened 2018-04-05
Added godot-mono scoop install option

murt:
I added the Mono version of Godot to scoop's extras bucket so here's a PR to put it up on the site if you think people would be interested in it.

Maybe should have some text about it being a different install as it might look like it should be installed after godot?


akien-mga:
Sorry for the late review. Adding it is fine, but it needs indeed to be made clearer that you need either `godot` or `godot-mono`, not both.

akien-mga:
Closing as stale.

PR #59, opened 2018-07-15
Custom 404 error page, Itch to download page

ghost:
Added a custom error page so you can always access the main nav.
Added a custom Godot error image and Itch logo.
Added Itch to download locations


djrm:
i would add an itch-download class with the itch red as background, that would look better.

PR #67, opened 2018-10-13
Grammar update, tense consistency and unnecessary comma

jcs224:

PR #68, opened 2018-10-13
Compound word

jcs224:
Object-oriented is a compound word.


LikeLakers2:
This PR has the same changes as #67, and these changes are also present in #69.

PR #73, opened 2019-01-07
Use HTTPS for links to docs.godotengine.org

merumelu:
By the way, all of the template files have their executable flag set. Is this intended?


akien-mga:
Thanks!

> By the way, all of the template files have their executable flag set. Is this intended?

No, it's probably just that someone modified them from Windows and we didn't set up git properly to prevent that.

PR #79, opened 2019-03-17
Labeled mono download as alpha

### LINKED ISSUE #50: Label Mono download as alpha
All.

**Issue description:**
The current download page doesn't make it clear that the Mono build is alpha software and does not reflect the overall stability of the non-Mono Godot 3.0.2.
screenshot_20180412_105144

It would be nice to make it crystal clear so that users know about it before they download and try to run it (especially since it might very well crash if they don't have the correct Mono SDK installed).

MHillier98:
Fixes #50 to reflect the state of the mono version to users downloading it.


Calinou:
Now that 3.2 is around the corner, should we still label it as Alpha? iOS and UWP exports are still missing, but it may not impact most users at this point.

akien-mga:
As mentioned in #50, enough time passed in the end so it's no longer necessary to label it as alpha on the download page.

PR #80, opened 2019-03-23
Update Link to Godot Developers forum

desttinghim:
I don't know why the URL changed, but it did. I'm fixing this so that those that want to use the forum are still able to find it without searching around.


Calinou:
The URL will have to be updated to https://godotforums.org/ as the forum is being moved to a more permanent location.

akien-mga:
Superseded by 0819656.

PR #87, opened 2019-05-23
Fix typos and improve static page contents

Calinou:
This updates some information (a compressed editor binary is closer to 25 MB than 20 MB nowadays) and improves grammar a bit.

This also adds a missing link to the GDScript introduction page on the Features page.


akien-mga:
Thanks!

PR #88, opened 2019-05-23
Make indentation consistent throughout the site files

Calinou:
This also adds an `.editorconfig` file which can automatically be used by text editors to configure indentation settings.


akien-mga:
Thanks!

PR #89, opened 2019-05-23
Optimize images losslessly

Calinou:
PNGs were optimized by running `oxipng -o6 --strip --zopfli` on them. SVGs were optimized by running `svgcleaner <input> <output>` several times on them; this gives better compression compared to doing it just once.

PS: Why are there duplicate images in `themes/godotengine/src/` and `themes/godotengine/assets/`? Which folder should new images be added to?


akien-mga:
Needs a rebase, otherwise seems fine.

> PS: Why are there duplicate images in `themes/godotengine/src/` and `themes/godotengine/assets/`? Which folder should new images be added to?

No idea. I think the CMS only uses the `assets` folder, so not sure what the `src` one is about.

akien-mga:
Closing for now, but it can be redone on the current `master` branch.

PR #90, opened 2019-07-23
Add a theme color

Calinou:
This color is used in various places such as in Chrome's navigation bar on Android, Discord embeds, etc.

The color chosen is identical to the CSS variable `--primary-color`.


akien-mga:
Thanks!

PR #92, opened 2019-08-04
Use the `font-display: swap` property with Google Fonts

Calinou:
This makes a fallback font display instantly, making the page readable as soon as it's loaded.


akien-mga:
Thanks!

PR #93, opened 2019-08-08
Fix some accessibility warnings reported by Lighthouse

Calinou:
See https://dequeuniversity.com/rules/axe/3.1/html-has-lang and https://dequeuniversity.com/rules/axe/3.1/link-name.


akien-mga:
Thanks!

PR #97, opened 2019-09-12
Add Moonwards as a gold sponsor

### LINKED ISSUE #96: Add Moonwards logo as Gold sponsor on homepage
[Moonwards](https://www.moonwards.com/) is a Gold sponsor since Sep 1, so their logo and a link to the website should be added to the homepage next to Gamblify's.

I tried to do it a week ago before going on holidays, but it looked bad due to our current CSS not being forgiving for their logo's aspect ratio.

SVG: https://www.moonwards.com/img/MoonwardsLogo2018.svg

So we need to tweak the CSS to better handle this kind of logo.

CC @Calinou @bojidar-bg @mrzapp.

Note that I'm on holidays, so feel free to handle merging a PR and updating the live website without me. Just be careful about the current git diff on the live instance due to the disabled Patreon module for php5 back compat.

Calinou:
The SVG was optimized using [svgcleaner](https://github.com/RazrFalcon/svgcleaner).

This closes #96.

I added @mrzapp as a co-author since I just followed his advice :slightly_smiling_face:


akien-mga:
Ah so just removing the base padding was enough, nice :)

PR #98, opened 2019-10-05
Avoid a redirect when clicking "More" in the navigation bar

Calinou:
This avoids a redirect when clicking "More" by going directly to the destination page. This makes the page load faster and decreases server load.


akien-mga:
Thanks!

PR #103, opened 2019-11-23
Add a link to the Godot Francophone website

xsellier:
Update the pages/user-groups.htm file to add a link to the
Godot Francophone Website


akien-mga:
As discussed on Discord, we should refactor this page to have a clearer distinction between country-based groups and international language-based groups (like this one).

But until this happens, having French-speaking groups listed under France is better than nothing I guess. We could also just use "France or French-speaking" as category.

xsellier:
Thank you @akien-mga 
I've updated the PR accordingly.

akien-mga:
Thanks!

PR #104, opened 2019-11-29
Add missing class

AinaSG:
Those 2 sponsors had the logo partially hidden and the wrong hover over animation.

### DIFF
--- a/themes/godotengine/layouts/home.htm
+++ b/themes/godotengine/layouts/home.htm
@@ -306,10 +306,10 @@ postPage = "{{ :slug }}"
       <a class="sponsor card base-padding" href="https://www.gamblify.com/" target="_blank">
         <img src="{{ 'assets/home/sponsors/gamblify.png' | theme }}" alt="Gamblify">
       </a>
-      <a class="sponsor card  base-padding" href="https://www.moonwards.com/" target="_blank">
+      <a class="sponsor card base-padding" href="https://www.moonwards.com/" target="_blank">
         <img src="{{ 'assets/home/sponsors/moonwards.png' | theme }}" alt="Moonwards">
       </a>
-      <a class="sponsor card  base-padding" href="https://www.imagecampus.edu.ar/" target="_blank">
+      <a class="sponsor card base-padding" href="https://www.imagecampus.edu.ar/" target="_blank">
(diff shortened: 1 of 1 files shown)


themes/godotengine/layouts/home.htm
```diff
@@ -306,10 +306,10 @@ <h2 class="text-center">Sponsored by</h2>
       <a class="sponsor card base-padding" href="https://www.gamblify.com/" target="_blank">
         <img src="{{ 'assets/home/sponsors/gamblify.png' | theme }}" alt="Gamblify">
       </a>
-      <a class="sponsor card" href="https://www.moonwards.com/" target="_blank">
+      <a class="sponsor card  base-padding" href="https://www.moonwards.com/" target="_blank">
```
Calinou: I know I'm nitpicking, but there should only be one space to separate each CSS class :slightly_smiling_face:

themes/godotengine/layouts/home.htm
```diff
@@ -306,10 +306,10 @@ <h2 class="text-center">Sponsored by</h2>
...
         <img src="{{ 'assets/home/sponsors/moonwards.png' | theme }}" alt="Moonwards">
       </a>
-      <a class="sponsor card" href="https://www.imagecampus.edu.ar/" target="_blank">
+      <a class="sponsor card  base-padding" href="https://www.imagecampus.edu.ar/" target="_blank">
```
Calinou: Same as above.

Calinou (changes requested):

themes/godotengine/layouts/home.htm
AinaSG: Upsies

Calinou:
It seems the current [Moonwards logo](https://godotengine.org/themes/godotengine/assets/home/sponsors/moonwards.png) is cropped by design (to make its height more reasonable), so I'll close this.

Also, [Image Campus is no longer sponsoring Godot](https://github.com/godotengine/godot-website/commit/2c3bea7c894c15804297f2e8a8ee57d7fbae8775).

AinaSG:
Then maybe it would be better crop the image? because the class with the animation and the margins is in all the other sponsors, right now it's not visible but if we add another sponsor without this one having the class, the animations break bad time.

PR #107, opened 2020-01-17
Add a link to the Complying with Licenses documentation page

### LINKED ISSUE #46: Add tentative license compliance instructions on the License page
As suggested [on reddit](https://www.reddit.com/r/godot/comments/83hxet/question_about_godot_licence_terms/), it would be a good idea to display further instructions on how to comply with the MIT license for Godot projects on the License page. There should be a disclaimer as those instructions should not be considered legal advice.

Some third-party libraries also require displaying the license text as well (or even an attribution notice, in the case of Freetype).

Calinou:
This closes #46.


akien-mga:
Thanks!

PR #109, opened 2020-01-17
Add a RSS feed icon that points to the RSS XML in the footer

### LINKED ISSUE #84: Add news subscription
**Issue description:**
Currently there is no way to make subscribe and receive news to email. It would be very nice to have this ability.

Calinou:
This closes #84.

## Preview

Social icons


akien-mga:
Thanks!

PR #112, opened 2020-01-17
Update some Yarn dependencies

Calinou:
Only dependencies that didn't introduce semver-major changes were updated. Further changes will be required to update all dependencies.

This makes it possible to install node-sass when using Node.js 10.x or later.


akien-mga:
Thanks!

PR #115, opened 2020-01-20
Add Docker config to setup and run the website

akien-mga:
This simplifies setting up the dev environment to run the website
(possibly with a database dump) and test changes before deploying
to the live instance.

---

First part of the work done by @mrzapp in https://github.com/mrzapp/godot-website, but rebased to squash commits together, improve the commit description, separate the Docker addition from the CMS plugins update, rebase on the master branch and fix line ending issues.

I'll make a second PR with the CMS plugins update, but we should likely test and validate the Docker setup first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,36 +7,39 @@ October instance.
 
 ### Dependencies
 
-- PHP 5 (version used in production)
-- [October](https://octobercms.com/)
-- Some database (MySQL is recommended)
+- [Docker](https://docker.com)
 - [Node.js](https://nodejs.org/) (6.x or newer)
   - If you cannot get a recent version from your distribution's repositories,
     try using [nvm](https://github.com/creationix/nvm).
 - [Yarn](https://yarnpkg.com/)
 
-### Installing October and the Godot theme & plugins
+### Running the site
 
-- Download and install [October](http://octobercms.com/).
-  - Note that you don't need to install Apache or nginx, as PHP has a built-in
-    development server. You can start it with `php -S localhost:8080` which
-    will host a server in the current directory.
-- If you have trouble with the installer, try to clone the repositories instead.
-  Running `composer install` then editing `config/database.php` with your
-  database credentials should get it up and running.
-- Once installed, clone this repository into your October directory (you'll have
-  to clone into an empty directory and then move it into the October directory).
-- Reassign the `active_theme` variable in `config/cms.php` to `godotengine`.
-- You may have some errors at this point, these are most likely due to plugins.
-  If an error is thrown, it should tell you which plugin caused it. Replace
-  `authorname` and `pluginname` with the plugin you wish to run the command on.
-  - Try installing the new plugin like so:
-    `php artisan plugin:install authorname.pluginname`
-  - Or refreshing the plugin like so:
-    `php artisan plugin:refresh authorname.pluginname`
-    (warning, this will erase all the database entries!)
-  - Plugins which might need these commands run on them are `RainLab.Blog` and
-    `PaulVonZimmerman.Patreon`
+- Clone this repo
+- Put a database dump (if you have one) into the `/docker/mariadb/init` folder
+    - Make sure that your script starts with the line `USE october;` and that the file extension is `.sql`
+- Run the `./docker/restart.sh` script (this will take a while the first time)
+- You might need to reinstall some plugins `/docker/php/install-plugin.sh author.name`
+    - Replace `author.name` with the names in the `/plugins/[author]/[name]` folders
+- See the website at [http://localhost:8080](http://localhost:8080)
+
+### Restoring a database
+
+- `mv /your/dump/backup-file.sql ./docker/mariadb/init`
+- `./docker/mariadb/bash.sh`
+- `cd /docker-entrypoint-initdb.d/`
+- `mysql < 000-setup.sql`
+- `mysql < backup-file.sql`
+
+### Interfacing with the Docker containers
+
+You can use the standard `docker exec -it godotengine-org--[php|mariadb] [command]` syntax or the following scripts:
+- `./docker/php/bash.sh`
+- `./docker/php/install-plugin.sh`
+- `./docker/php/log.sh`
+- `./docker/mariadb/bash.sh`
+- `./docker/mariadb/log.sh`
+- `./docker/mariadb/mysql.sh`
 
 ### Setting up the theme
 
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -1,0 +1,36 @@
+FROM php:apache
+
+# Settings
+WORKDIR /var/www/html
+ENV COMPOSER_ALLOW_SUPERUSER=1
+
+# Install server dependencies
+RUN apt-get update &&\
+    apt-get install -y git unzip vim zlib1g-dev libzip-dev libpng-dev
+
+# Install PHP extensions
+RUN docker-php-ext-install mysqli pdo pdo_mysql zip gd
+
+# Enable PHP modules
+RUN a2enmod rewrite
+
+# Install composer
+RUN curl -sS https://getcomposer.org/installer -o composer-setup.php &&\
+    php composer-setup.php --install-dir=/usr/local/bin --filename=composer
+
+# Install OctoberCMS
+RUN git init . &&\
+    git remote add origin https://github.com/octobercms/october.git &&\
+    git pull origin master &&\
+    composer install
+
+# Config
+RUN sed -i "s/'host'      => 'localhost'/'host'      => 'mariadb'/g" config/database.php
+RUN sed -i "s/'database'  => 'database'/'database'  => 'october'/g" config/database.php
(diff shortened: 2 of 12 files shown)


docker/php/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN a2enmod rewrite
+
+# Install composer
+RUN curl -sS https://getcomposer.org/installer -o composer-setup.php &&\
+    php composer-setup.php --install-dir=/usr/local/bin --filename=composer
+
+# Install OctoberCMS
+RUN git init . &&\
+    git remote add origin https://github.com/octobercms/october.git &&\
+    git pull origin master &&\
```
akien-mga: I might look into making this configurable, so that we can test (and deploy) stable releases instead of the `master` branch. This would also make it possible to test the Docker setup against the current CMS and plugins versions that we have deployed on godotengine.org (OctoberCMS 419).

README.md
```diff
@@ -7,36 +7,39 @@ October instance.
 
 ### Dependencies
 
-- PHP 5 (version used in production)
-- [October](https://octobercms.com/)
-- Some database (MySQL is recommended)
+- [Docker](https://docker.com)
```
akien-mga, line 10: `docker-compose` is also needed (not installed automatically at least when installing `docker` on my distro).

README.md
```diff
@@ -7,36 +7,39 @@ October instance.
...
-    `php artisan plugin:refresh authorname.pluginname`
-    (warning, this will erase all the database entries!)
-  - Plugins which might need these commands run on them are `RainLab.Blog` and
-    `PaulVonZimmerman.Patreon`
+### Running the site
+
+- Clone this repo
+- Put a database dump (if you have one) into the `/docker/mariadb/init` folder
+    - Make sure that your script starts with the line `USE october;` and that the file extension is `.sql`
+- Run the `./docker/restart.sh` script (this will take a while the first time)
```
akien-mga, line 21: This works fine, but I have nothing on `http://localhost:8080`. It's not clear to me if any other step should be done to access the website on my host browser.

README.md
```diff
@@ -7,36 +7,39 @@ October instance.
...
+- You might need to reinstall some plugins `/docker/php/install-plugin.sh author.name`
```
akien-mga, line 22: Any plugin that I try to install gives me:
```
$ sudo docker/php/install-plugin.sh rainlab.blog

In Connection.php line 664:
                                                                                                                                               
  SQLSTATE[HY000] [2002] No such file or directory (SQL: select `created_at` from `system_plugin_versions` order by `created_at` asc limit 1)  
                                                                                                                                               

In PDOConnection.php line 47:
                                                    
  SQLSTATE[HY000] [2002] No such file or directory  
                                                    

In PDOConnection.php line 43:
                                                    
  SQLSTATE[HY000] [2002] No such file or directory  
                                                    
```

README.md
akien-mga: From reading docs on Docker Compose, it should be work out of the box.. I tried to disable my firewall, but I now get this error:
```
$ sudo ./restart.sh 
Removing network docker_godotengine-org--network
WARNING: Network docker_godotengine-org--network not found.
Creating network "docker_godotengine-org--network" with the default driver
ERROR: Failed to Setup IP tables: Unable to enable SKIP DNAT rule:  (iptables failed: iptables --wait -t nat -I DOCKER -i br-88e89f792815 -j RETURN: iptables: No chain/target/match by that name.
 (exit status 1))
```

akien-mga: `systemctl restart docker` fixed it.

akien-mga: I guess I'm not using the database dump properly as I get this error on localhost:
```
SQLSTATE[HY000] [2002] No such file or directory (SQL: select * from `rainlab_blog_categories` where `slug` = news order by `nest_left` asc limit 1)

/var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php line 664
```

mrzapp:
No such file or directory doesn't sound like a database error, but rather disk I/O, even though the PDO is throwing the exception. Are you somehow using SQLite instead of MariaDB, perhaps? Or is there a plugins folder missing?

akien-mga:
> Are you somehow using SQLite instead of MariaDB, perhaps?

I have both SQLite and MariaDB installed on the host, but I would expect it to use MariaDB as configured in Docker Compose? I'm not clear on how Docker for web services works exactly to be honest, I've only ever used Docker for build containers :)

> Or is there a plugins folder missing?

I have an up to date git clone, so all plugins should be there. To be sure that it's not due to one of the missing plugins update in this PR, I'm actually testing with the branch from #116 to have the same state you had eventually.

akien-mga:
I might have misunderstood what's needed for the database dump setup though.

I took the mysql dump generated by TuxFamily daily, and placed it as `docker/mariadb/init/godotengine_wp.sql` (with permissions fixed to `644`).
I added `USE october;` as the first line in the SQL script.

Then ran:
```
cd docker
sudo ./restart.sh
```

mrzapp:
The database dump should be restored like this, just to be clear:  

- `mv /your/dump/backup-file.sql ./docker/mariadb/init`
- `./docker/mariadb/bash.sh`
- `cd /docker-entrypoint-initdb.d/`
- `mysql < 000-setup.sql`
- `mysql < backup-file.sql`

Make sure the SQL script starts with `USE october;`

It doesn't matter what's installed on your host system other than docker and docker-compose. The needed software, such as MariaDB, is installed inside the container anyway.

akien-mga:
I've tried that too, though it was not clear to me if this was a pre-requisite to getting things running, as it comes after the main howto. I'll add a clarification once I get it working.

Do I need to do anything specific after `mysql < backup-file.sql`? It doesn't seem to suffice to make things work on `localhost:8080`.

akien-mga:
The error I get in October itself makes it seem like it did find the DB, but maybe it's not compatible with latest OctoberCMS / updated rainlab.blog plugin?
https://github.com/godotengine/godot-website/pull/115#discussion_r368430603

mrzapp:
About your issue exposing the PHP container on the local network, the error you mentioned [here](https://github.com/godotengine/godot-website/pull/115#discussion_r368427985) seems to have a solution in [this](https://github.com/moby/moby/issues/16816#issuecomment-179717327) discussion.

About the error in your log, it could just be that rainlab.blog isn't doing schema checks properly before attempting to access data, and might not have an impact on the site running, or could fix itself after the first time writing to the table.

In any case, I suggest you focus on getting a response from your PHP container before dealing with other issues. I'll be here to help, if it still won't behave after that.

akien-mga:
> About your issue exposing the PHP container on the local network, the error you mentioned here seems to have a solution in this discussion.

Thanks. As I mentioned in the next comment, restarting the docker service was sufficient to solve it. So I do have access to the PHP container now, but the DB issues prevent using it:

Screenshot_20200121_102558

On http://localhost:8080/backend I get:
Screenshot_20200121_102759

mrzapp:
I'd spin it up again to see if I can reproduce this error, but I need another DB dump for that. Could you send one to me on matrix again? I can't seem to find the credentials for the DB admin interface anywhere.

akien-mga:
Looks like I had a good first comment here: https://github.com/godotengine/godot-website/pull/115#discussion_r368412777

I hardcoded the October version from when you initially wrote the docker setup, and now it works:
```diff
diff --git a/docker/php/Dockerfile b/docker/php/Dockerfile
index 99be7ca..74d9121 100644
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -21,7 +21,8 @@ RUN curl -sS https://getcomposer.org/installer -o composer-setup.php &&\
 # Install OctoberCMS
 RUN git init . &&\
     git remote add origin https://github.com/octobercms/october.git &&\
-    git pull origin master &&\
+    git fetch origin &&\
+    git checkout v1.0.458 -b october-1.0.458 &&\
     composer install
 
 # Config
```

It can now properly read the DB, and does complain about some plugins, but that's expected as per the setup instructions.

I suggest we make this configurable, and pinned to a specific version by default, to ensure that we get reproducible containers (with a way to override locally for contributors who want to test upgrades).

mrzapp:
Alright, cool! That means they have introduced breaking changes somewhere, but if it's not in their stable branch anyway, we could just hang on to this version for the time being.

akien-mga:
Alright, I got things to work locally now.

I pushed an update to make the OctoberCMS version configurable. I wanted to default to version 419 which we have on the live instance, but it can't be installed with current composer due to composer/composer#7665...

So instead I'm defaulting to the stable version @mrzapp was initially testing against. To make it fully work, some plugins updates and indeed necessary (basically what is currently in #116), but I'll merge this in a second step.

If someone manages to make version 419 install properly with composer, the current state of the PR should be usable to get the website running locally with the same config as the live instance.

akien-mga:
Added a last patch to ensure that file ownership is set back to the calling user when using `install-plugin.sh`.

akien-mga:
Thanks a lot for the setup (and help debugging) @mrzapp!

PR #116, opened 2020-01-20
WIP: Update OctoberCMS plugins to match version 458 (Aug 2019)

akien-mga:
Second part to integrate @mrzapp's work to upgrade the CMS and plugins for PHP 7 support.

Depends on and includes #115, so it should be rebased once that PR is merged.

There was no explanations about the plugins and home layout updates in the original commits, so I used my best guess. The Docker files seem designed to install OctoberCMS from its master branch, so I assumed that the plugins update was done to match the then latest OctoberCMS version, which is 458 (459 was released a couple of weeks later in September 2019).

We'll have to check if the update is good to deploy as is against latest OctoberCMS 464 or if another round of updates is needed (if so, we should document in the process how this was done).


akien-mga:
Once this is reviewed, approved and merged, what's not clear to me is how to deploy this on the live instance of the blog on godotengine.org.
I clearly lack webdev experience and I'm not familiar with a Docker workflow for these things.

The plugins update will simply be a matter of `git pull`, but updating the CMS itself needs to be handled differently. Should I somehow copy the working updated CMS from the Docker container?

mrzapp:
Docker can't be part of the deployment workflow on tuxfamily's servers with their limitations.

I think, after you pulled the latest from the website's Git repo, that you need to copy the files from the latest distribution of OctoberCMS manually into the root dir, overwriting old ones.

This could be automated via a bash script later on, just for convenience.

In any case, a little downtime might happen, which is common when upgrading a major version of anything. If we can coordinate it, I'll be glad to be on stand-by on matrix while you're doing this.

akien-mga:
I rebased on #115 and updated the Rainlab Blog plugin again.

Then I had errors on the News and Article pages due to the use `{{ post.user }}` which seems undefined for some reason. I couldn't figure out why, so I removed those for now in 9a154d8.

Then I can log in fine in the backend using my credentials, but it seems that the CMS still identifies itself as version 419, even though I should be running version 458:
Screenshot_20200121_131219

I assume that it takes the number from the database dump, but that means that it's not able to update the DB with new info? Or do we miss a step in the install process?

To see what it does, I tried to let the backend do the upgrade to version 464 by itself, but it fails with an error which hints at some permission issues:
```
"ZipArchive::extractTo(): Permission denied" on line 74 of /var/www/html/vendor/october/rain/src/Filesystem/Zip.php
```

YuriSizov:
I don't know how applicable this is to the production version, but as I has been recently trying to spin up the setup locally, I have to share that the pinned version of October in the `docker-compose.yml` file (`v1.0.458`) has an internal problem with its `wikimedia/composer-merge-plugin`. 

This plugin is referred to by the `dev-master` version, so the latest revision of it is installed. At some point this plugin started to require PHP to be `7.2` or higher. Which is fine in theory, as the docker is installed with PHP `7.4`. At the same time, though, October is overriding your installed version value with the `config.platform.php` property of `composer.json` to read as `7.0`. This applies to every tagged version of October, though it has been changed to `7.2` in master.

This causes this error:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for wikimedia/composer-merge-plugin dev-master -> satisfiable by wikimedia/composer-merge-plugin[dev-master].
    - wikimedia/composer-merge-plugin dev-master requires php >=7.2.0 -> your PHP version (7.4.9) overridden by "config.platform.php" version (7.0) does not satisfy that requirement.
```

The issue persists until `v1.0.465` of October, where they pin the version of `wikimedia/composer-merge-plugin` to the one where there is no conflict. So, to conclude, maybe we should update October to `v1.0.465`.

Calinou:
This was superseded by the CMS and plugin upgrade we carried out last week, closing. We now run October 1.0.469.

YuriSizov:
Docker files need to reflect the new versions, though 🙃

PR #124, opened 2020-02-23
Add instant.page to speed up page loading

Calinou:
instant.page prefetches pages when the user hovers links for more than 65 milliseconds. This can be used to make navigation feel significantly faster.

Note that this will have no effect in Firefox when using uBlock Origin, as it forcibly disables prefetching.

See also https://github.com/godotengine/godot-docs/pull/3197.


akien-mga:
I don't have a strong opinion on this, not sure if it's actually a good feature or not. WDYT @bojidar-bg?

mhilbrunner:
We've added this to the docs, IMO while not groundbreaking it's a nice small improvement.

bojidar-bg:
If we are fine with it in the docs, it should be okay in the main website.

akien-mga:
Alright, let's merge once rebased then. @Calinou

Calinou:
Rebased. I also updated instant.page to the latest available version in the process.

akien-mga:
Thanks!

PR #136, opened 2020-05-23
Use standard case for headings instead of uppercase

Calinou:
Standard case is easier to read compared to uppercase. It tends to look better overall as well. For the record, I did this for the game titles in the [Godot 3.2 games showcase](https://www.youtube.com/watch?v=UEDEIksGEjQ).

The main page title now uses a non-breaking space between "waited" and "for" to prevent leaving just "for" on a new line.

## Preview

<details>
<summary>Various screenshots with the new case</summary>

</details>


akien-mga:
Thanks!

PR #137, opened 2020-05-23
Tweak the text about headless versus server builds on the Download page

Calinou:
See https://github.com/godotengine/godot-docs/pull/3557#issuecomment-633089196. Thanks @follower for the suggestion :slightly_smiling_face:


akien-mga:
Thanks!

Issue #83, opened 2019-04-06
Make VR support more visible on the website


BastiaanOlij:
I'm often getting reactions of people who are surprised Godot has VR support, and even more surprised when I tell them we've had support for well over a year, probably two :)

Maybe one of the tiles on the front page could mention AR and VR support?
Yeah that does make 7 which is a problem :)

MHillier98:
An 'easy' solution could be to add both VR and something like Visual Shaders, or some other major feature to that list?

On similar note, I think it could be worth adding links from that features to the docs or something? It would be nice for newcomers to be able to click directly to more information about the features that are being displayed. Worth a new issue?

BastiaanOlij:
Bump for this, seems to be easier to add in the current setup on the features page. Happy to do a PR myself if someone could point me in the right direction (or is this a change that needs to be done in the database)

@Calinou , akien said you might be able to point me in the right direction.

Calinou:
@BastiaanOlij This should be a matter of modifying this page: https://github.com/godotengine/godot-website/blob/master/themes/godotengine/layouts/features.htm

Add new images in this folder (not `src` which no longer seems to be used): https://github.com/godotengine/godot-website/tree/master/themes/godotengine/assets/features

Calinou:
This was done in #138, closing.

PR #138, opened 2020-05-25
Adding section about XR support to features

BastiaanOlij:
Not sure how to test this but first attempt at implementing #83


akien-mga:
Thanks!

PR #139, opened 2020-05-31
Apply the theme's line height change to ordered lists

Calinou:
Articles like <https://godotengine.org/article/godot-editor-running-web-browser> use ordered lists, but the line height was previously only applied to paragraphs and unordered lists.

### Before

image

### After

image


akien-mga:
Thanks!

PR #146, opened 2020-06-11
Store Patreon plugin inside the repository

Calinou:
This removes the need to initialize a (broken) submodule URL to get a local copy of the website running. I'm not sure if it's worth using submodules just for a single plugin, so I think vendoring it makes more sense here.

The plugin was cloned from https://github.com/pcvonz/oc-patreongoalstatus-plugin
(commit 5d685ddc9bbb522d668f7906185f24b1319da3b3).

Note that it still requires the `patreon/patreon` Composer dependency to be installed. You can do this by running `composer install` inside the plugin folder. Should we also vendor it so people don't have to do it manually? None of the other plugins require this (they already seem to have vendored dependencies).


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "plugins/paulvonzimmerman/patreon"]
```
akien-mga: Why is this added if we're not using a submodule in the end?

Calinou: I think I accidentally added it. I'll remove it.

akien-mga:
Thanks!

PR #147, opened 2020-06-19
Use a text link for the asset library navbar link instead of an icon

### LINKED ISSUE #74: Proposal - Replace Padlock/Shopping Basket with text link to Asset Lib.
https://github.com/godotengine/godot-website/blob/7ad427d60a678c3e4fe75ec58ea62073afca18a0/themes/godotengine/partials/header.htm#L28

Honestly. Neither a Padlock, Nor a shopping basket are intuitive. What are we protecting? What are we selling/buying? It took me until yesterday, since the creation of the asset lib, to realize that that is what the icon is for. I usually just google it.

Placing it next to the Donation button is also confusing. Imho it should sit next to Community, before More.

Calinou:
This closes #74.


akien-mga:
Thanks!

PR #151, opened 2020-06-30
Don't capitalize page titles automatically

Calinou:
The CSS capitalize property doesn't take stop words into account, so it's probably not a good idea to rely on it. Moreover, we prefer capitalizing only the first word instead of doing so for every word.


akien-mga:
Thanks!

PR #152, opened 2020-07-08
Use HTTPS URLs everywhere possible to avoid redirects

Calinou:
This makes pages load slightly faster as there is no more redirect.


akien-mga:
Thanks!

PR #153, opened 2020-07-09
Add press kit assets.

fire:
Added the contents of the zip from https://github.com/godotengine/godot/issues/8514#issuecomment-333914851

For another PR:

*   Logo usage guidelines
*   Engine screenshots
*   Game/showcase screenshots
*   Info text
*   Color codes


Calinou:
We probably want a ZIP to be made available somewhere so people don't have to download each file individually. This means it will have to be generated at deployment time.

fire:
Can you describe a brief design for that? @Calinou

Calinou:
@fire I think @akien-mga will create the ZIP when deploying the website.

fire:
@akien-mga Friendly ping

akien-mga:
I don't really see much added value in this PR as it stands.

There's a README + files, but nothing that is actually accessible to users who visit the website unless they know exactly what URL to type.

The assets are tons of variations on `logo.svg` which haven't been approved yet, and they're hard to review here in PR form.
If the only thing we provide is variations on the logo, they need to be official variations that we're happy to see used in the wild. Personally I'm not fond of monochrome black and monochrome white versions (for what I could see by opening some of the assets on GitHub).

So instead of a PR that adds a dump of files somewhere in the repo, I think that if we want a press kit, we need to define:
- What it should contain
- Have discussions/mockups for the art assets that it should contain, so we can agree on them beforehand
- Decide where to host those assets and how to make them discoverable
- Etc.

fire:
I agree.

The original post was from 2017 https://github.com/godotengine/godot/issues/8514#issuecomment-333914851

Do you want to go through the process of the press kit creation?

* We need to return to discussion to the art assets that the press kit should contain
* Create a page for the press kit to make them discoverable. @Calinou Could probably think of some ideas

YuriSizov:
Superseded by #292.

PR #157, opened 2020-07-21
Remove unused theme source files

Calinou:
We no longer rely on a build process to work on the theme files. This is easier for contributors, and the added complexity isn't worth the trouble for our use cases.

@akien-mga Is [`themes/godotengine/deploy.js`](https://github.com/godotengine/godot-website/blob/11278db816911775d520d545c1319e451349bd93/themes/godotengine/deploy.js) still used? It refers to an `assets/packed` folder which will no longer be generated by Webpack here.


akien-mga:
Needs a rebase as I merged some dependabot stuff before this.

I have no clue what all this was added for so if we don't need it, then yes I'm happy to remove it.

Calinou:
@akien-mga Rebased.

> I have no clue what all this was added for so if we don't need it, then yes I'm happy to remove it.

These files were added in the 3.0 website redesign. It seems we've since started to modify the generated files directly, making the source files obsolete.

(Moreover, CSS preprocessors aren't as useful as they used to be, since we can now use native CSS variables.)

akien-mga:
Thanks!

PR #158, opened 2020-07-21
Fix reflows in Home and Features pages by specifying aspect ratios

Calinou:
Modern browsers will reserve the size automatically as long as the `width` and `height` attributes are used to specify the *aspect ratio*, even if CSS is later used to resize the image.

This prevents reflows while the page is loading, and therefore improves perceived performance.

Backgrounds were also added to opaque images so they have a visible placeholder while loading.


akien-mga:
Thanks!

PR #162, opened 2020-07-27
Fix unordered list of scoop install commands

### LINKED ISSUE #145: Update the bullet points on the Download page's Instructions section
Windows 10, Google Chrome 83.0.4103.97

**Issue description:**
The bullet points currently being used inconsistently, with the first `<pre>` tag having a point, and the second not having one.

Personally, I think it would look better to not use them in that section, but would also be OK with just adding another bullet point to the second `<pre>` tag for "scoop install godot".

**Screenshots:**
Current section:
image

Proposed change:
image

MHillier98:
Closes #145 - there are two scoop install commands in the same unordered list element.

### DIFF
--- a/themes/godotengine/pages/download/windows.htm
+++ b/themes/godotengine/pages/download/windows.htm
@@ -41,6 +41,8 @@ is_hidden = 0
   <ul>
     <li>
       <pre><code>scoop bucket add extras</code></pre>
+    </li>
+    <li>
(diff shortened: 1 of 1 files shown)


Calinou:
Screenshot comparison:

## Before

image

## After

image

Calinou (approved):
Thanks!

PR #165, opened 2020-08-15
Fix reflows when loading the Community and Download pages

Calinou:
Follow-up to https://github.com/godotengine/godot-website/pull/158.

Specifying the image widths and heights in advance makes it possible for the browser to reserve the space while the page is loading.

This improves the user experience while decreasing CPU usage.


akien-mga:
Thanks!

PR #167, opened 2020-08-22
Add syntax highlighter to code blocks in articles

YuriSizov:
This PR adds a syntax highlighter in the form of Highlight.js to the articles. Highlight.js has a benefit of having a dedicated ruleset for GDScript maintained by our contributors. The change is retroactive and affects all existing articles/news. 

I picked a dark theme called `Obsidian` for this, but there are a lot of other options if you think that something else would fit better. I naturally lean towards dark themes for code, but there may be a reason to use a lighter theme instead.

image

-----
The retroactivity of this PR has a side effect. All code blocks that don't require any highlighting should be marked as `plaintext`. If the language is not specified, Highlight.js will try to "detect" it. For example, this is Ruby to it:

image

/cc @akien-mga @Calinou


themes/godotengine/layouts/article.htm
```diff
@@ -94,6 +104,11 @@
   }
 </style>
 
+<link rel="stylesheet" href="{{ 'assets/css/highlight.obsidian.min.css' | theme }}">
+<script src="{{ 'assets/js/highlight.min.js' | theme }}"></script>
+<script src="{{ 'assets/js/highlight.gdscript.min.js' | theme }}"></script>
```
Calinou: I think you could use the `defer` attribute to speed up loading a bit:

```suggestion
<script defer src="{{ 'assets/js/highlight.min.js' | theme }}"></script>
<script defer src="{{ 'assets/js/highlight.gdscript.min.js' | theme }}"></script>
```

Other than that, this PR looks good to me. Nice work :slightly_smiling_face:

YuriSizov: Sure! Init script had to be adjusted too of course.

akien-mga:
Thanks!

PR #164, opened 2020-08-06
Improve styling of articles for better readability

YuriSizov:
Whenever I look at one of the news on the main website I notice at least three things that bother me:
* Articles take very little horizontal space, making them harder to comprehend;
* The empty space around the text doesn't anchor your eyesight very well: by the middle of any article the text kind of floats somewhere in the middle of the page;
* `<code>` tag is not styled beyond using monospaced font.

I've tried to address all of the concerns mentioned. As I am not familiar with this project, I've tried to reduce the impact of the changes by packing everything into the article template, but I think some of the changes can be applied to other pages as well. If these changes are undesired or require some adjustments in the implementation, please let me know.

Here are some comparisons:
**Top of the page**:
[Before](https://user-images.githubusercontent.com/11782833/89535604-4efc2080-d7ff-11ea-8182-9abf225c8045.png) | [After](https://user-images.githubusercontent.com/11782833/89535651-5fac9680-d7ff-11ea-9905-b396f881bc3b.png)

**Middle of the article**:
[Before](https://user-images.githubusercontent.com/11782833/89535753-84087300-d7ff-11ea-9e15-e59a97eaf10d.png) | [After](https://user-images.githubusercontent.com/11782833/89535764-89fe5400-d7ff-11ea-86a0-3016c106a339.png)

**More code tags**:
[Before](https://user-images.githubusercontent.com/11782833/89535812-9b476080-d7ff-11ea-9f48-d4ecd1f17234.png) | [After](https://user-images.githubusercontent.com/11782833/89535827-a00c1480-d7ff-11ea-961f-cf01e102e6fe.png)

**Footer**:
[Before](https://user-images.githubusercontent.com/11782833/89535840-a8fce600-d7ff-11ea-83c5-b9b5d046b4f9.png) | [After](https://user-images.githubusercontent.com/11782833/89535854-ae5a3080-d7ff-11ea-8298-1ad73b0394af.png)

PS. Due to a rather complex setup I wasn't able to test it live, but I took care to test it as best as possible with Chrome devtools.

PR #169, opened 2020-08-22
Make the footer dark on every page

YuriSizov:
Makes the footer styles implemented in #164 apply to every page. I had to adjust the look of the sponsors block on the home page a bit, but otherwise it just fitted right in.

For example, here's the [home page](https://user-images.githubusercontent.com/11782833/90960889-8defce80-e4ad-11ea-9143-6200cac79cff.png).

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -249,9 +249,10 @@ ul li {
 
 
 footer {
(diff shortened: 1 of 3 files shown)


Calinou (approved):

akien-mga:
Thanks!

PR #173, opened 2020-08-26
Clarify that the GitHub repository is for bug reports only

Calinou:
- Also link to the Godot proposals repository in the description.


akien-mga:
Thanks!

PR #175, opened 2020-08-26
Update the Events pages for the current situation

Calinou:
This also fixes duplicate or broken IDs on the past events page.


akien-mga:
Thanks!

PR #176, opened 2020-08-27
Fix inline code styling

YuriSizov:
I think I misjudged some CSS features while doing #167, sorry for that.

This should restore the styling for inline code (depicted below) while not breaking code blocks:
image


Calinou (approved):

akien-mga:
Thanks!

PR #178, opened 2020-09-18
Allow internationalization of the website

### LINKED ISSUE #47: Internationalize the official web site
Could you please support translations for the official web site like the godot editor ?
If the web site speaks native languages that would make godot more attractive to new people.

YuriSizov:
Now, this is a big one. Closes #47. Read up on that issue if you want to know how this has come to be.

This PR unlocks the ability to internationalize our website to a number of languages. Main changes are as follows:

* Added the `RainLab.Translate` plugin for OctoberCMS, allowing for translations;
* Created a custom `GodotEngine.Utility` plugin to extend the functionality of the Translate plugin to suit our needs;
* Covered almost all existing pages making texts translatable (with a couple of exceptions, see below);
* Partially translated those pages to Russian (mostly to test things, but the translations are proper);
* Added a locale switcher to the footer.

**Commits are purposefully unsquashed.** I hope this would allow to check the code and the layouts easier. There is a lot to check and I've tried to make commits as concise and related to one subject as possible. Be advised, though, that though I've tried to address all of the technical details in the first few commits, some issues arose later on. I've tried to keep things separate as much as possible still.

### How does it work?

The system is two-fold. RainLab.Translate provides the low level tools for translations to be applied to the texts. Unfortunately, this plugin operates on an internal database while we prefer to keep translations in files so that they can be maintained via Weblate. The Translate plugin supports files, but only CSV (via import) and YAML (via scanning). We rely on YAML files as the only reasonable way to automate things.

However, actual translations are kept in PO files, as is the longstanding tradition when translating Godot resources. Those PO files are generated from the actual template code of the website using GodotEngine.Utility. They can then be moved anywhere outside of the system, edited and re-added. As RainLab.Translate is not able to handle PO files directly, we use them to generate YAML files. The Translate plugin happily scans them and adds translated strings. The rest is controlled by it.

To enforce a language on a page the path of a URL is used. For example, `https://godotengine.org/features` can be translated into Russian by changing the URL to `https://godotengine.org/ru/features`. On the subsequent requests to the website the locale is remembered and kept, even if the URL is not containing the locale code. These links can be cached by search engines. We may need to think about how we want to handle that. At the very least, canonical meta-tags are required.

As mentioned before, there is now the locale selector at the bottom of each page. So manual editing of URLs is not needed.

### How to translate the website?

Compared to the approach described in the linked issue things got a bit simpler. Text still needs to be wrapped in a particular construct, however it's just one function call. There is a small extra detail here: as we are using PHP, single quotation marks are pretty powerless, and double quotation marks are "magical". This is utilized in the following manner:

* For single lines of texts (buttons, headers, etc) use a single quotation mark: `{{ TR('Some text') }}`;
* For multiline statements, especially containing markup and formatting, use a double quotation mark: `{{ TR("Some text") }}`.

Here's the example from the actual code:

```twig
      <h2>{{ TR('The game engine you waited&nbsp;for.') }}</h2>

      <p>
        {{ TR("
        Godot provides a huge set of common tools, so you can just focus on
        making your game without reinventing the wheel.
        ") }}
      </p>
```

Technically, this is it. Though one has to remember to escape single and double quotes in the corresponding strings (with a backslash).

```twig
    {{ TR("
    By joining Godot communities, you agree to follow Godot's
    <a href=\"/code-of-conduct\">Code of Conduct</a>.
    ") }}
```

### How to work with PO files?

To generate PO files a locale must exist. To add a new locale use the admin panel (*Translate > Manage Languages*). This step cannot be automated as it must be performed on the actual database. To generate the necessary files, use the following command: `php artisan godotengine:addi18n es` (where `es` is the code for the locale; it must match with the code entered in the admin panel).

```shell
root@28925d7870fb:/var/www/html# php artisan godotengine:addi18n es
Adding "es" translation.
Writing translation PO file themes/godotengine/i18n/po/messages.es.po.
Writing translation YAML file themes/godotengine/i18n/es.yaml.
Updating Theme configuration.
Successfully added "es" translation.
```

PO files are updated using another command: `php artisan godotengine:extracti18n` (no arguments). This command scans Twig templates for the calls to the aforementioned `TR` function and extracts all calls using string literals for arguments. Firstly, a base file is generated anew. Then this file is merged using `msgmerge` with the PO file for each locale. That's it, the resulting PO files can be edited. `msgmerge` tries to fuzzy map similar messages too, so small changes to the original lines can be propagated through translations without any loss of values.

```shell
root@28925d7870fb:/var/www/html# php artisan godotengine:extracti18n
Starting message extraction.
Scanning Layout files.
Scanning Page files.
Scanning Partial files.
Merging extracted messages with the current base file.
Writing extracted messages to the base file.
Updating locale-specific files.
........ done.
Successfully updated "ru" translation; file themes/godotengine/i18n/po/messages.ru.po has been written.
Finished message extraction.
```

After the files have been edited by the translators, the last command must be called (`php artisan godotengine:updatei18n`) to generate YAML files and update the database. The database is updated with *purging*, anything that was stored in it will be overridden.

```shell
root@28925d7870fb:/var/www/html# php artisan godotengine:updatei18n
Starting translation update.
Successfully updated "ru" translation; file themes/godotengine/i18n/ru.yaml has been written.
Updating translation database.
Flushing CMS cache.
Finished translation update.
```

Obviously, the database must be updated on the server. However, YAML files should be generated beforehand. Currently, they are stored in the repository. Theoretically, they can be ignored and only exist on the server. However you decide.

### Caveats

There are several things that I must describe about this implementation.

1. The website is not made to be translated, and it shows. There are several cases where layouts are broken by strings too long. Message collisions are also an issue. For example, the word *Donate* used to denote the page describing ways to donate to the project should be translated into Russian with two words (that is if we use proper formal language and not modern slang). But in some cases we could make it shorter to fit the need of the UI. For example, in the footer it can be the full phrase *"Поддержать проект"*, but on the *More* pages we could use just the first word to make the tabs more compact:

image

But we cannot achieve that, as the underlying English variation is the same. We may be able to introduce some context information to solve this later. I can imagine a second argument for the `TR` function and some special comments in the PO file which can be parsed wherever this data is needed.

And continuing with these pages, here's the example of the layout breaking:

image

I'm pushing the PR as is, though, to get reviews on the progress so far and get help from other contributors on fixing those. After all, the localizations can be in the *beta* state for some time, I think.

2. Speaking of message collisions, string ids used in YAML files must be relatively short. The Translate plugin limits them to 250 characters, but this limitation doesn't work properly and breaks for me at about 200 characters. Plus, the Translate plugin doesn't work properly with multiline strings and string with an HTML markup. So I had to make a workaround to generate ids with our Utility plugin in a way that obscures all those issues from the Translate plugin. I chose the following strategy to do so:

* Generate the initial ID in the same way as the Translate plugin;
* Crop it to 150 characters, which seems to work fine locally;
* To increase uniqueness, append the length of the remaining string to the end.

Oh, and if the current locale is English none of this is done, the Translate plugin isn't even used. The strings are just returned as is by the `TR` function.

3. There are many texts with markup. I took a look at the way we handle it in other cases. Unfortunately, [we don't use HTML anywhere](https://hosted.weblate.org/translate/godot-engine/godot-docs/ru/?q=state:%3E=translated), instead relying on other formatting methods. But it would be detrimental to break those texts into pieces for translators to compile back together and bend the languages to fit the patterns. I've researched a bit some ways to sanitize HTML without breaking it completely in PHP, but ultimately it makes more sense to do this before importing the translations than at the runtime.

Thankfully, Weblate seems to have some validation tools, including [HTML sanitation](https://docs.weblate.org/en/latest/user/checks.html#unsafe-html). This needs to be set up and tested. Because the implementation of this translation system absolutely trusts that the texts are sanitized to the degree required and **absolutely allows for XSS** to be injected if we are not careful.

Another option would be to move from HTML entirely... At least for those texts. But that's a lot more work, obviously.

4. I've skipped translating the past events page, because I frankly see little sense in doing this. This page just seems to need some reorganization. Maybe it shouldn't be a static page at all. But anyway, future events can be covered in the translation wrapping when the time comes.

I did, however, cover the legal and formal documents, like License and CoC, with the markup, though I did not translate them. I'm not sure those documents should be translated by just anyone, the license in particular, but that's up to you. I did have to adjust *a lot* of layouts across many pages, those included. This was required to add translations wrappings nicely. But I could've broken some important texts. This must be validated, I just don't know how except by hand. And it's a lot of work.

But ultimately, the changes to the layout should be welcome, as they make those files much more readable. Some work is needed to clean them up at some point, though, I believe.

5. The locale switcher is using the base path of the pages to generate links, which works fine for our case, because our paths map one to one with the URLs. But this may not always be the case. I had implemented a much hackier solution with the JS at first, and it worked fine. But I later removed it in favor of a somewhat built-in solution with the base paths. Still, we may want to fall back to the JS solution, if this current approach is considered a worse idea.

Here it is, by the way:

image

6. The titles of the most pages are translated automagically, because they match the strings already present in the system. But for the downloads page, at least, I've had to come up with a way to change it to a translated variant via some use of placeholders:

```swift
{% put head_title %}{{ TR('Download') ~ ' | Linux' }}{% endput %}
```

This can be applied to any page that requires adjustments. But, like I've said, it mostly *just works*:

```swift
<title>Godot Engine - {% placeholder head_title default %}{{ TR(this.page.title) }}{% endplaceholder %}</title>
```

-----

I am sure something important must've skipped my mind. So please ask any questions you have.

PS. I have to say, it's super freaking cool to [see the website translated](https://user-images.githubusercontent.com/11782833/93628162-c7edbb00-f9ee-11ea-9620-dd9ba5f2f550.png).

/cc @akien-mga @Calinou @Geequlim


Calinou:
@pycbouh I was thinking about this PR and ways to simplify it, since right now, it adds quite a lot of complexity. Why not use [Weblate's YAML support](https://docs.weblate.org/en/latest/formats.html#yaml-files) directly so we don't have to convert between PO and YAML? While YAML is a less advanced format, we most likely won't need things like pluralization in the website translations.

YuriSizov:
> Why not use Weblate's YAML support directly so we don't have to convert between PO and YAML?

I didn't consider that, but at a glance this doesn't seem like an optimal solution. How well formed do the keys need to be? Is there a size limit to a key? Can they be merged in a similar manner to `msgmerge`?

PO allows us to preserve the untranslated content as is, which removes the need to sanitize it, doesn't introduce an artificial size limit, increases uniqueness of keys. It also seems that Weblate supports a specific format of YAML files, so we'd have to convert from October i18n YAML to Ruby YAML. Just to clarify, YAML is only a part of this picture because it's the only feasible way to interface with October. If we could use PO files directly, I'd prefer us to do that, but alas.

I don't think that this PRs implementation is that complex, but I understand your hesitation as a sole maintainer of our web ecosystem. For my part, I hope that the plugin won't require maintanance at all, it's pretty straightforward. I can document the code better, if required. Using it shouldn't be a problem neither.

YuriSizov:
This PR is very outdated and hard to rebase unless there is very good reason for it. As the main concern is complexity, and I don't think we'll get anything better in that regard with October, I'll close it for now. We better move to a better CMS, I guess.

This can be salvaged if needs be.

PR #179, opened 2020-09-26
Link to the Troubleshooting page on the Download page

Calinou:
This helps people find it more easily.

### DIFF
--- a/themes/godotengine/layouts/download.htm
+++ b/themes/godotengine/layouts/download.htm
@@ -153,6 +153,11 @@ description = "Download layout"
           <h4>Instructions</h4>
           <ul>
             <li>Extract and run. Godot is self-contained and does not require installation.</li>
+            <li>
+              If you run into an issue, check the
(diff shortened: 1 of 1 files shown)


themes/godotengine/layouts/download.htm
```diff
@@ -153,6 +153,11 @@ <h4>Requirements</h4>
           <h4>Instructions</h4>
           <ul>
             <li>Extract and run. Godot is self-contained and does not require installation.</li>
+            <li>
+              If you run into an issue, check the
+              <a href="https://docs.godotengine.org/en/latest/about/troubleshooting.html">Troubleshooting</a>
```
akien-mga: This should likely link to `stable` (but first I need to merge `master` in `stable` for the docs).

akien-mga: https://docs.godotengine.org/en/stable/about/troubleshooting.html should be valid in half an hour or so, so you can use that. (I synced the docs 2 days ago but `stable` failed building for a non-deterministic reason while `3.2` succeeded with the same code.)

akien-mga:
Thanks!

PR #184, opened 2020-10-05
Add comment blocks for syntax highlighting support in Visual Studio Code

Calinou:
Using the OctoberCMS Template Language add-on, it's possible to have proper syntax highlighting for October templates in Visual Studio Code. However, due to a limitation of this add-on, we need to add empty comment blocks at the top of every template.

See https://marketplace.visualstudio.com/items?itemName=dqsully.octobercms-template-language.


akien-mga:
Thanks!

PR #185, opened 2020-10-05
Use a single code block for Docker commands in the README

Calinou:
This makes it easier to copy all the lines at once for later editing.


akien-mga:
Thanks!

PR #186, opened 2020-10-05
Document browser support for the website

Calinou:
This also makes the `#godotengine-atelier` IRC link clickable.


akien-mga:
Thanks!

PR #190, opened 2020-10-16
Remove custom 404 page as it breaks the CMS after upgrading to 1.0.469

Calinou:
I don't know why, but we have to remove it so all pages can load correctly (without displaying the 404 page's contents in place of the Community and More pages).


akien-mga:
Thanks!

PR #195, opened 2020-10-22
Remove Flexbox-based CSS grid fallback

Calinou:
[All evergreen browsers support the CSS grid module now.](https://caniuse.com/css-grid)


akien-mga:
Thanks!

PR #201, opened 2020-11-09
Add a community map on the User groups page

Calinou:
The map is powered by [Leaflet.js](https://leafletjs.com/) and [Mapbox](https://www.mapbox.com/)'s tiles API. Mapbox's free tier is limited to 50,000 page loads per month, which should be fine given the community map is placed on a 3rd-level page (*Home > Community > User Groups*).

User groups are now stored in a PHP array located in the front matter. This makes the [godotengine/community-map](https://github.com/godotengine/community-map) repository obsolete. While not as ubiquitous as JSON, the PHP array format is more commonly known compared to the custom format we were using before. (We could still move the list of communities to a JSON file if needed.)

When adding a new country, make sure to add its GPS coordinates to the `countryCoordinates` JavaScript object in `user-groups.htm`. This is required to place the online community marker correctly.

## Preview

<details>

Screenshot_2020-11-09 User Groups
</details>


akien-mga (approved):

akien-mga:
> User groups are now stored in a PHP array located in the front matter. This makes the godotengine/community-map repository obsolete. While not as ubiquitous as JSON, the PHP array format is more commonly known compared to the custom format we were using before. (We could still move the list of communities to a JSON file if needed.)

I liked the idea to have data sourced from a dedicated repository where people can directly PR their details, but I guess we can start with the embedded PHP for now and see if there's a need to move it out eventually.

froquede:
Just out of curiosity, why use the mapbox tile server instead of OSM?

Calinou:
> Just out of curiosity, why use the mapbox tile server instead of OSM?

Does OpenStreetMap still offer a freely usable tile server? Last time I checked (2 months ago), you had to use a third-party tile provider.

froquede:
Yes, it does! https://wiki.openstreetmap.org/wiki/Tile_servers
And leaflet supports the OSM multiple subdomains for speeding up concurrent requests by default:
https://switch2osm.org/using-tiles/getting-started-with-leaflet/
https://leafletjs.com/reference-1.7.1.html#tilelayer

Calinou:
@roquef Feel free to open a pull request to switch to an OSM tile server :slightly_smiling_face: 

Ideally, it would be better if no API token was required to make local development easier. Right now, you need to specify a token to get tiles and this token is scoped to the domain name (to avoid hitting the rate limit too soon due to someone else using it).

PR #204, opened 2020-11-28
Add spacing between Events and Community blocks on Community page

name-here:
It seems weird that Events and Tutorials would be grouped together, so I added space between them to match the space between User groups and Events.  Apologies if this isn't desired, or I'm doing this wrong.


akien-mga:
Thanks!

Issue #205, opened 2020-11-30
License tutorial page 404s


LaffertyDev:
**Operating system, Browser, Browser Version:**
Kubuntu 20.04, Firefox, 83.0

**Issue description:**
Click on "Complying with Licenses". Expect to be taken to [good link](https://docs.godotengine.org/en/stable/tutorials/legal/complying_with_licenses.html) but taken to [bad link](https://docs.godotengine.org/en/latest/tutorials/legal/complying_with_licenses.html)

**Screenshots (greatly appreciated if it’s a style issue)**
n/a

Page: https://godotengine.org/license

deralmas:
Since its respective PR has been merged, shouldn't this issue be closed?

Calinou:
> Since its respective PR has been merged, shouldn't this issue be closed?

Indeed, thanks for the reminder :slightly_smiling_face:

PR #206, opened 2020-11-30
Point license tutorial link to new directory

LaffertyDev:
This corrects an issue #205 where the tutorial page 404s


YuriSizov:
Actually, we can and probably should still point to the latest. It's just been moved here: https://docs.godotengine.org/en/latest/about/complying_with_licenses.html

LaffertyDev:
No problem, updated link again. Went the path-of-least resistance and didn't notice the file had actually moved.

Calinou (approved):
Thanks!

PR #208, opened 2020-12-08
Add platform icons to the showcase list for easier recognition

Calinou:
Games that are supported on highly sought-after platforms (like consoles) can now be seen immediately on the list.

## Preview

Screenshot_2020-12-08 Showcase


akien-mga:
Thanks!

PR #213, opened 2020-12-17
Prevent the More tab navigation from taking too much space from the title

Calinou:
This is needed for https://github.com/godotengine/godot-website/pull/178 and the Governance page I'll submit in the future.

### Before

Tabs overflowing

### After

Tabs not overflowing


akien-mga:
Thanks!

PR #220, opened 2020-12-25
Increase the number of blog posts per page to 25

Calinou:
This makes it easier to browse through the blog archives while requiring less page loads. I tested this change locally by creating 16 blog posts; they now all appear on the same page.

The previous number of posts per page was 15.


akien-mga:
Thanks!

PR #222, opened 2021-01-04
Replace instant.page with Barba.js for faster subsequent page loads

Calinou:
Advantages:

- Turns the page into an actual single-page application, which provides more performance benefit compared to just prefetching the page before loading it. The website still works without JavaScript as this is just a progressive enhancement.
- Works on Firefox with uBlock Origin enabled.

This also fixes the `<body>` tag never being opened and improves the HTML formatting to make the raw source more readable.

Internal links now consistently use absolute paths instead of external links.

I tested this change locally, including with articles that contain highlighted code blocks.

To get an idea of how fast this can be, visit a Barba-enabled site such as [yours truly's website](https://hugo.pro/) :slightly_smiling_face:


akien-mga:
Thanks!

PR #236, opened 2021-01-07
Decrease the font weight for inline code

Calinou:
The inherited font weight (400) will now be used instead of 600.

### Before

Font weight 600

### After

Font weight 400


akien-mga (approved):

akien-mga:
Thanks!

PR #251, opened 2021-01-16
Remove link to Better Collada exporter as it's unmaintained

Calinou:
This also clarifies that Godot isn't limited to importing ESCN scenes, as it can import glTF 2.0 and OBJ.

See discussion in https://github.com/godotengine/godot-proposals/issues/2130 before merging.


akien-mga:
Thanks!

PR #254, opened 2021-01-20
Update brew command for OSX download page

le-michael:
`brew cask <command>` is deprecated and `brew <command> --cask` is preferred.  Currently running the command throws an error `Calling brew cask install is disabled! Use brew install [--cask] instead.`

https://github.com/Homebrew/discussions/discussions/340

The change updates the command to use the `--cask` flag.


Calinou (approved):

PR #264, opened 2021-01-21
Adjust Russia's coordinates in User Groups

YuriSizov:
Russia is a tad more to the north, actually.

image


Calinou (approved):
Thanks!

I don't remember where I got the broken coordinates from, but it's been surprisingly difficult for me to find a centralized list of country center coordinates.

danilaboni:
thanks for fixing it

PR #267, opened 2021-01-21
Change Mapbox tile server to OSM

froquede:
As discussed in #201, it changes the tile server used by leaflet.js from Mapbox to OSM in the user groups map.


Calinou (approved):
Tested locally, it works. Thanks!

PR #275, opened 2021-01-26
Features: Remove Perforce and PlasticSCM as they're seldom tested

Calinou:
This partially addresses https://github.com/godotengine/godot-docs/issues/4584.

### DIFF
--- a/themes/godotengine/layouts/features.htm
+++ b/themes/godotengine/layouts/features.htm
@@ -265,7 +265,7 @@ description = "Features layout"
     <p>Godot is designed from the ground up for smooth teamwork.</p>
 
     <ul>
-      <li><strong>Friendly filesystem usage</strong> that works great with version control systems like Git, Subversion, Mercurial, PlasticSCM, Perforce, you name it.</li>
+      <li><strong>Friendly filesystem usage</strong> that works great with version control systems like Git, Subversion, Mercurial, you name it.</li>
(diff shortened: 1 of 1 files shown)


akien-mga:
I think the referenced issue might be a misunderstanding. This statement here doesn't mean that Godot *supports* Perforce integration, but that its text-based and well compartmentalized scenes play nice with version control systems as a whole. The enumeration is just to match names that game developers might be familiar with (some might know they use Perforce at work without even knowing that it's a "version control system").

akien-mga:
What should we do with this change? I'm OK merging it but as mentioned above, I don't think it's necessary (and many in the gamedev do use Plastic or Perforce so it's useful to let them understand that Godot will play nice with their favorite VCS - even without native integration).

Calinou:
@akien-mga It seems Perforce itself is poorly supported due to read-only mode not playing well with Godot:

> There does not seem to be a perforce add-on, nor does it seem to natively work very well with perforce (read-only causes a lot of errors in the engine) The features page as well as the File system page mention Perforce by name as something that works well with Godot.
> —https://github.com/godotengine/godot-docs/issues/4584

Read-only mode (and file locking in general) are one of the big reasons game developers like to use Perforce instead of Git. Without proper read-only support in the Godot editor, I don't think we can claim to actually support Perforce.

akien-mga:
Alright, makes sense.

PR #276, opened 2021-01-27
Add Chinese Community on community map.

wangdiandao:
It add the information of Chinese Community on this page.


akien-mga:
Added another commit to fix a missing `],` and put it in alphabetical order (so between Canada and Croatia).

akien-mga:
I checked the GPS coordinates and they seem to match Beijing, which I guess makes sense. Alternatively, OpenStreetMap suggests `35.0000, 105.0000` as "random point in the middle of China" which could be used if we need to distinguish between China-wide community and a Beijing regional one. But it's probably good as is for now.

akien-mga:
Thanks!

PR #282, opened 2021-02-02
Update user-groups.htm

AnatolyBelyaev:


themes/godotengine/pages/user-groups.htm
```diff
@@ -178,6 +178,7 @@
         'links' => [
           ['title' => 'VKontakte', 'url' => 'https://vk.com/godot_engine'],
           ['title' => 'Telegram', 'url' => 'https://t.me/godot_engine'],
+          ['title' => 'VK', 'url' => 'https://vk.com/game_constructors'],
```
YuriSizov, line 181: Since this is not a part of `Godot Russia`, but is an independent community, you need to add it the same way `Skanersoft` is added below, with a dedicated record. Also, please use the full name of `VKontakte`, as show above.

YuriSizov (changes requested):

PR #283, opened 2021-02-03
Fix Open Graph image meta tag on articles

Calinou:
The `name` attribute was duplicated, which meant the `twitter:image` value usually "won" over the `og:image` value.


akien-mga (approved):

akien-mga:
Thanks!

PR #285, opened 2021-02-04
Fix Barba.js not taking external Godot websites into account

### LINKED ISSUE #269: Downloading PDF from the GodotCon news yield unexpected results
Windows 10, Chrome 87.0.4280.141
Also confirmed in Firefox Nightly by akien.

**Issue description:**
Pressing the *Download this PDF* link at the bottom of the [GodotCon 2021 schedule news](https://godotengine.org/article/online-godotcon-2021-schedule) doesn't correctly download a PDF document. Instead, it is loaded and immediately replaced in the browser history by the page you are currently on.

image

I think this may be an issue with the plugin we use for faster page loading (#222). Maybe also the same issue as #228.

**Screenshots (greatly appreciated if it’s a style issue)**

Calinou:
This removes the need for individual `data-barba-prevent` attributes.

This closes #269.

### DIFF
--- a/themes/godotengine/pages/privacy-policy.htm
+++ b/themes/godotengine/pages/privacy-policy.htm
@@ -29,9 +29,9 @@ is_hidden = 0
     <li>you visit the Godot web site (<a href="/">godotengine.org</a>), hosted by <a href="https://www.tuxfamily.org/">TuxFamily</a> (see the TuxFamily privacy policy: <a href="https://faq.tuxfamily.org/Privacy/En">https://faq.tuxfamily.org/Privacy/En</a>);</li>
     <li>you access the Asset Library in the Godot editor application to browse or download assets;</li>
     <li>you sign up as a Godot Patreon (<a href="https://patreon.com/godotengine">https://patreon.com/godotengine</a>) or otherwise donate to Godot (<a href="/donate">https://godotengine.org/donate</a>);</li>
-    <li>you sign up and log in to an account on the Godot Chat platform (<a href="https://chat.godotengine.org" data-barba-prevent>https://chat.godotengine.org</a>), and use it to participate in private and public discussions;</li>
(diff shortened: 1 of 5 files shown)


themes/godotengine/pages/privacy-policy.htm
```diff
@@ -29,9 +29,9 @@ <h3 class="title">The Information We Collect</h3>
...
     <li>you access the Asset Library in the Godot editor application to browse or download assets;</li>
     <li>you sign up as a Godot Patreon (<a href="https://patreon.com/godotengine">https://patreon.com/godotengine</a>) or otherwise donate to Godot (<a href="/donate">https://godotengine.org/donate</a>);</li>
-    <li>you sign up and log in to an account on the Godot Chat platform (<a href="https://chat.godotengine.org" data-barba-prevent>https://chat.godotengine.org</a>), and use it to participate in private and public discussions;</li>
-    <li>you sign up and log in to an account on Godot's Q&amp;A site (<a href="https://godotengine.org/qa" data-barba-prevent>https://godotengine.org/qa</a>), and use it to ask or answer questions;</li>
-    <li>you sign up and log in to an account on Godot's Asset Library (<a href="https://godotengine.org/asset-library" data-barba-prevent>https://godotengine.org/asset-library</a>), and use it to submit content;</li>
+    <li>you sign up and log in to an account on the Godot Chat platform (<a href="https://chat.godotengine.org">https://chat.godotengine.org</a>), and use it to participate in private and public discussions;</li>
```
Calinou, line 32: `data-barba-prevent` should be unnecessary here as it's not the same subdomain. If I'm reading its source code right, Barba.js checks for the whole hostname to check whether the website is external, not just the domain name.

YuriSizov:
So we'll just have to maintain a list of exceptions?

Calinou:
> So we'll just have to maintain a list of exceptions?

Yes :slightly_smiling_face:

PR #298, opened 2021-03-10
Improve pagination display logic for news articles

Calinou:
Follow-up to https://github.com/godotengine/godot-website/pull/297.

Inspired by GitHub's pagination display logic:

- Display 5 page links close to the current page, but always display the first two and last two pages. Add ellipses if at least one page is omitted between two ranges.
- Disable Previous and Next links instead of hiding them entirely. This makes the pagination bar size more predictable for the user.

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -438,7 +438,8 @@ footer a {
   font-size: 0;
 }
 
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #304, opened 2021-03-17
Add a warning for logged in users when viewing an unpublished article

Calinou:
Unpublished articles can be previewed by logged in users, but they can't be seen by non-logged in users.

The link brings you to the CMS backend's relevant post and tab so that you can quickly check the **Published** checkbox and set a publication date in the past.

## Preview

image


akien-mga:
Thanks!

PR #308, opened 2021-03-23
Add logo previews on press kit page

DeeJayLSP:
It would be great if we could visualize a preview of each logo just scrolling by the page


Calinou (approved):

Merged manually as https://github.com/godotengine/godot-website/commit/d6173f6d3a12cd8de42942de7fa1075f56f1f8bd with some changes:

- Logos are now clickable.
- Logo width is specified in the HTML to avoid page reflows during loading.
- Logos are lazy-loaded for better initial load performance.

PR #313, opened 2021-03-29
Update user-groups.htm. Fix mistake from last commit

bogdasar1985:
Mistake came out with that https://github.com/godotengine/godot-website/pull/312/commits/ebf9f8e4b117e8014fcca470e994bb0fff08327d commit. Fixed

### DIFF
--- a/themes/godotengine/pages/user-groups.htm
+++ b/themes/godotengine/pages/user-groups.htm
@@ -192,21 +192,15 @@ function onStart() {
         ],
       ],
       [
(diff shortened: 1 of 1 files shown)


Calinou:
As discussed on the Godot contributors chat, we decided to just update the Skanersoft user group name to Godot Engine Russia but keep the current Discord invite link. @pycbouh confirmed that there was no reason to change the invite link.

Thanks for the contribution nonetheless :slightly_smiling_face:

PR #317, opened 2021-04-06
Add instructions for installing Godot using MacPorts

jasonliu--:
Greetings Godot devs,

I am the creator and maintainer of the Godot package for the [MacPorts](https://www.macports.org) package management system. Godot Engine has been available as an official MacPorts package since May of 2020. Since then, I have sent a couple of e-mails to [contact@godotengine.org](mailto:contact@godotengine.org) (which were replied to by Remi Verschelde), to have some instructions added for MacPorts to the macOS [Download page](https://godotengine.org/download/osx), but nothing appears to have gotten added. So I thought that I would try submitting a PR myself.

### DIFF
--- a/themes/godotengine/pages/download/osx.htm
+++ b/themes/godotengine/pages/download/osx.htm
@@ -36,7 +36,7 @@ is_hidden = 0
     You can also get Godot with <a href="https://brew.sh/">Homebrew</a> or <a href="https://www.macports.org">MacPorts</a>.
     Note that the version available on Homebrew isn't code-signed either.
     Software released through MacPorts does not need to be code-signed to run;
-    this means it is not affected by Gatekeeper restrictions.
+    this means Godot on MacPorts is not affected by Gatekeeper restrictions.
   </p>
   <ul>
     <li><pre><code>brew install --cask godot</code></pre></li>
--- a/themes/godotengine/pages/download/osx.htm
+++ b/themes/godotengine/pages/download/osx.htm
@@ -33,10 +33,11 @@ is_hidden = 0
     <a href="https://store.steampowered.com/app/404790">Godot from Steam</a> to work around this.
   </p>
   <p>
-    You can also get Godot with <a href="https://brew.sh/">Homebrew</a>.
+    You can also get Godot with <a href="https://brew.sh/">Homebrew</a> or <a href="https://www.macports.org">MacPorts</a>.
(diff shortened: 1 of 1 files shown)


themes/godotengine/pages/download/osx.htm
```diff
@@ -33,10 +33,11 @@
     <a href="https://store.steampowered.com/app/404790">Godot from Steam</a> to work around this.
   </p>
   <p>
-    You can also get Godot with <a href="https://brew.sh/">Homebrew</a>.
+    You can also get Godot with <a href="https://brew.sh/">Homebrew</a> or <a href="https://www.macports.org">MacPorts</a>.
     Note that the version available on Homebrew isn't code-signed either.
```
Calinou, line 37: If MacPorts releases don't run out of the box due to Gatekeeper restrictions, I suggest amending this notice to avoid confusion:

```suggestion
    Note that the versions available on Homebrew and MacPorts aren't code-signed either.
```

Official macOS releases will be signed and notarized starting from 3.3, but I don't think this will be done retroactively for 3.2.x as this would require making new binaries for them.

jasonliu--: No, that change isn't necessary. Software released through MacPorts runs out of the box and is not affected by Gatekeeper restrictions, and doesn't need to be code-signed/notarized.

I actually thought about changing the sentence to something like: "Note that the version available on Homebrew isn't code-signed either; software released through MacPorts does not need to be code-signed, and so is not affected." but I decided to leave in its original form for my PR submission.

Calinou, line 37:
```suggestion
    Note that the version available on Homebrew isn't code-signed either.
    Software released through MacPorts does not need to be code-signed to run;
    this means it is not affected by Gatekeeper restrictions.
```

themes/godotengine/pages/download/osx.htm
```diff
@@ -33,10 +33,13 @@
...
+    Software released through MacPorts does not need to be code-signed to run;
+    this means it is not affected by Gatekeeper restrictions.
```
Calinou:
```suggestion
    this means Godot on MacPorts is not affected by Gatekeeper restrictions.
```

Calinou (approved):

jasonliu--:
Awesome, thanks so much!

PR #318, opened 2021-04-08
Remove style attribute in /themes/godotengine/pages/error.htm

MrBrain295:
Remove depreciated style attribute in /themes/godotengine/pages/error.htm.


Calinou:
The `style` attribute is here by design. It's not bad practice to use the `style` attribute for things you don't need to reuse several times in the same page.

Also, using IDs in CSS tends to be frowned upon nowadays, since you can only use each ID once per page.

PR #323, opened 2021-04-14
Press kit: Change previews background color to match website theme

DeeJayLSP:
Except for outlined icon since we need to demonstrate the clash


Calinou (approved):
Thanks!

PR #332, opened 2021-05-15
Add `security.txt` file for standardized security issue reporting

Calinou:
See https://securitytxt.org/ and https://github.com/godotengine/godot/pull/45701#issuecomment-799897418 (which I had missed).

As per the specification, the file should be updated once every year to make sure it doesn't go stale.

I've checked on the production instance and `.well-known/security.txt` can be accessed fine. (I removed the file until this PR is merged.)

**Edit:** The presence and validity of the `security.txt` file can be tested here: https://gotsecuritytxt.com/query/godotengine.org


akien-mga (approved):

akien-mga:
Thanks!

PR #333, opened 2021-05-18
Update Download page for 3.3.1-stable

akien-mga:


YuriSizov:
Wouldn't it make more sense to utilize CMS capibilities for that? Would likely be more manageable 🙃

Calinou:
> Wouldn't it make more sense to utilize CMS capibilities for that? Would likely be more manageable :upside_down_face:

I don't think October has a good interface for that, but we can make better use of the `stable_version` variable and add a `stable_version_date` variable that accepts an ISO 8601 date and formats it nicely.

PR #334, opened 2021-05-19
Update user-groups.htm

wangdiandao:
Add Godot QQ group in china.


Calinou:
@RealSYWW If this QQ group is not created by the same people who created https://godotengine.top/, you need to put this user group in a new community instead of adding a link to an existing community. This is because we want to be clear about *who* manages a certain community.

wangdiandao:
I created both of them

akien-mga (approved):

akien-mga:
Thanks!

PR #340, opened 2021-05-27
Update IRC community link to point to Libera Chat

Calinou:
Most of the activity has moved to Libera Chat now. I tested the web chat URL and it works as expected :slightly_smiling_face:


akien-mga (approved):

akien-mga:
Thanks!

PR #341, opened 2021-06-05
Update Godot Mono prerequisites: .NET SDK

### LINKED ISSUE #321: Update Godot Mono prereqs on download page for 3.2.3+
N/A, content issue.

**Issue description:**
The prerequisites listed for Godot Mono on https://godotengine.org/download for any platform are:

> image

Now you just need a .NET [Core] SDK. This is what the docs describe:
https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#prerequisites

> image

31:
Update the Godot Mono prerequisites on https://godotengine.org/download to make it match up with the docs:
https://docs.godotengine.org/en/stable/getting_started/scripting/c_sharp/c_sharp_basics.html#prerequisites

@neikeq mentioned in https://github.com/godotengine/godot/issues/41734#issuecomment-690764782 (Sep 10, 2020) that disabling non-`dotnet CLI` build tools is being considered, so funneling more people towards the .NET SDK seems to me like a good idea even if VS Build Tools and Mono SDK are working right now. (Although based on people getting help in the Godot C# Discord, they do seem to cause issues pretty often, and switching over to `dotnet CLI` normally clears them up.)

(The .NET SDK corresponds to the `dotnet CLI` Build Tool option in the Editor Settings.)

Fixes https://github.com/godotengine/godot-website/issues/321

### DIFF
--- a/themes/godotengine/layouts/download.htm
+++ b/themes/godotengine/layouts/download.htm
@@ -134,7 +134,7 @@ description = "Download layout"
           <ul>
             <li>OpenGL 2.1 / OpenGL ES 2.0 compatible hardware</li>
             <li>
-              <strong>For the Mono version:</strong> MSBuild<br />(from <a href="https://visualstudio.microsoft.com/downloads/?q=build+tools">Visual Studio Build Tools</a> or the <a href="https://www.mono-project.com/download/stable/">Mono SDK</a>)
+              <strong>For the Mono version:</strong> <a href="https://dotnet.microsoft.com/download">.NET SDK</a> or the <a href="https://www.mono-project.com/download/stable/">Mono SDK</a>
             </li>
           </ul>
         </div>


akien-mga (approved):

akien-mga:
This seems good but I'm not sure it's a good idea to merge as long as `dotnet CLI` is not the default option, so this might have to wait. CC @neikeq

neikeq:
> This seems good but I'm not sure it's a good idea to merge as long as `dotnet CLI` is not the default option, so this might have to wait. CC @neikeq 

`dotnet CLI` has been the default for quite some time. As long as Godot can find it, that is.

neikeq:
Problem with this is lack of .NET Sdk for 32-bit Linux.
Godot 4.1 will drop C# support on 32-bit Linux, at least until the .NET Core supports it. But we still support it on 3.x and 4.0.

31:
> Problem with this is lack of .NET Sdk for 32-bit Linux.

Interesting, didn't know about this reason to use Mono. 👍

It doesn't look like Microsoft has plans to produce linux-x86 builds for the .NET runtime: https://github.com/dotnet/runtime/issues/7335. (Although that might not be the only thread about it.) After that, getting a linux-x86 SDK might be a separate effort.

For the time being, would it make sense to list out the three options on the download page and just recommend the .NET SDK over the others?

van800:
Listing just .NET SDK and mono would be enough then.

mhilbrunner:
The commits should be squashed into one :)

31:
I've added back in the Mono SDK link, so it would look like this according to my browser:

image

Does that work?

31:
> The commits should be squashed into one :)

Is this a new style? In repo history, it looks like GitHub's squash merge feature has been used pretty frequently, so it wouldn't matter. I was planning to not force push to make the history of this PR clear. Happy to squash it myself though.

van800:
@neikeq please merge this, there are people on discord every week puzzled by VS build tools requirement.

akien-mga:
Thanks!

PR #347, opened 2021-06-23
Remove cacheroute plugin, breaks live instance

akien-mga:
We've had it deleted locally on the live instance for a while, but it would be better
to make the Git version match what we do use in production.


YuriSizov:
Oh, that's good to know. I think this one also prevented me from setting up a local development version on a new PC.

Calinou (approved):

PR #350, opened 2021-07-07
Optimize press kit images using `oxipng -o6 --strip --zopfli`

Calinou:
This also optimizes the LITSLINK sponsor image.


akien-mga:
Thanks!

PR #352, opened 2021-07-17
Add Godoter Planet chinese user group to the community map

timothyqiu:
This is an online community.

The code is not tested locally because I'm currently having trouble running docker-compose :( But I believe this won't break anything :P


Calinou (approved):
Thanks!

PR #360, opened 2021-08-27
Link to the GitHub Releases mirror on the Download page

Calinou:
This is useful for users having issues with slow downloads from TuxFamily's servers.

## Preview

image

### DIFF
--- a/themes/godotengine/layouts/download.htm
+++ b/themes/godotengine/layouts/download.htm
@@ -153,9 +153,10 @@ description = "Download layout"
           {% placeholder instructions %}
         </div>
 
-        <div class="footer base-padding">
+        <div class="footer base-padding" style="line-height: 1.6; font-size: 85%">
           You can find previous releases on the
-          <a href="https://downloads.tuxfamily.org/godotengine">download repository</a>.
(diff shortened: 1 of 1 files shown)


YuriSizov:
So, will you add it to the mirror list for export templates as well?

Calinou:
> So, will you add it to the mirror list for export templates as well?

I can do that independently of this PR, as mirrorlist files aren't stored in this repository.

**Edit:** I don't have permission to edit the files on the server, so Akien will have to do that once he gets back from vacation.

mhilbrunner:
Wouldn't it make sense to make Github the default? I assume their latency/bandwith/CDN capabilities are usually going to be superior, or at least not worse.

Calinou:
> Wouldn't it make sense to make Github the default? I assume their latency/bandwith/CDN capabilities are usually going to be superior, or at least not worse.

I think it's better to change this incrementally at first to see if anyone runs into problems. We can make GitHub Releases the default when 4.0 gets close to release.

mhilbrunner:
Makes sense.

mhilbrunner (approved):

Issue #353, opened 2021-07-17
Tabs in news page has white padding when viewed in narrow window


timothyqiu:
**Operating system, Browser, Browser Version:**
macOS 11.4 / Chrome 91.0.4472.164 & Safari 14.1.1

**Issue description:**

When the browser window is narrow, there are unexpected white padding below & right to the news page's tabs.

**Screenshots (greatly appreciated if it’s a style issue)**
![padding](https://user-images.githubusercontent.com/372476/126042660-0f63ffd9-5d07-4685-9554-4549cf53f880.png)

YuriSizov:
That looks like the space for scrollbars. What's your window resolution?

timothyqiu:
It's about 880x700 in the screenshot.

Calinou:
We should probably disable `overflow-y` on that tab container.

bits-by-brandon:
@Calinou confirmed that https://github.com/godotengine/godot-website/pull/364 fixed this. Should we close this issue?

YuriSizov:
Yes, good call. It wasn't linked properly, but it seems to be fixed.

PR #364, opened 2021-09-13
Fix tabs having extra padding in mobile view

avanwinkle:
**#353 Tabs in news page has white padding when viewed in narrow window**

This PR changes the tab container on mobile view to use `overflow: auto` instead of `overflow: scroll`. The auto behavior will trigger scrolling when necessary (highly unlikely with only two tabs), but removes the scrollbar padding otherwise.

Before:
Screen Shot 2021-09-13 at 3 13 25 PM

After:
Screen Shot 2021-09-13 at 3 13 34 PM

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -928,7 +928,7 @@ pre > code {
   }
 
   .tabs {
-    overflow: scroll;
+    overflow: auto;
   }
 
   .author .date {


Calinou:
Make sure to test this change on the "More" pages too, such as the Governance page. There are much more tabs on "More" pages, so horizontal scrolling must remain available in case of an overflow.

avanwinkle:
Verified that it works as expected, with a horizontal scrollbar appearing on the Governance page as necessary.

Screen Shot 2021-09-13 at 8 26 25 PM

The changes are tested using the dev tools, I'm unable to build and run the website locally to verify. I'll keep investigating and possibly raise a question in the dev chat if I can't get past the composer dependency mismatch.

Calinou (approved):
Thanks!

Issue #210, opened 2020-12-08
Add ability to filter by platform in the showcase list of games made with godot section


Maciej-Gorczyca:
**Operating system, Browser, Browser Version:**
Win 10 Chrome

**Issue description:**

In a showcase allow users to filter games and applications by different platforms. If someone is a mobile developer and wants to see what games are made with it on mobile they could just select android and iOS to be displayed and ignored the rest. Or if you make games on PC you could just select some of pc platforms like Steam, Itch. 

Maybe consider adding filter for 2d and 3d games as well as those tend to be very different breeds and some people may only be interested in seeing one or the other. 

Finally separate programs from games as it's own section. Programs made with godot are something unique really and would be nice to give is separate showcase section as they are their own "selling" point of Godot development. 

**Screenshots (greatly appreciated if it’s a style issue)**
N/A

Calinou:
Right now, the showcase is purely static, which means this would have to be done client-side with some JavaScript code. It might be a good opportunity to use [Alpine.js](https://github.com/alpinejs/alpine)…

a2937:
Hello. May I offer my input? I'd like to redo the showcase collections such that windows , macos, linux, ios, and android variables were part of an array instead so that these categories can be added as classes to the project card. From there I would make a drop down that would be called from a quick script at the bottom of the page. No external files needed. And here's some additional reading material on how I want to do things.

[Apply all classes](https://stackoverflow.com/questions/15998587/automatically-add-class-when-listing-all-jekyll-posts)

[Hide based on Javascript](https://stackoverflow.com/questions/18414384/hide-element-by-class-in-pure-javascript)

coppolaemilio:
This one was implemented in this PR: https://github.com/godotengine/godot-website/commit/ea554479ca50a3337ab7c442aefc352a95de4256

Thanks!

PR #371, opened 2021-10-02
feat: Filter for the showcase items.

dploeger:
This features a filter for the showcase items as described in #210 

The filter looks like this:

image

It can filter the title, type, platform and tags by using the filter UI and also by using the query params like this:

    /showcase?title=dung&tags[]=narrative&tags[]=rpg#showcase

I guess, we should ask the developers what tags they would like to use. This is currently just a bunch of tags I thought would fit.


YuriSizov:
Needs to be redone for the new static website (though @coppolaemilio already plans to work on it next).

PR #373, opened 2021-10-15
CI: Relax pinned version for actions/checkout and setup-php

akien-mga:


.github/workflows/ci.yml
```diff
@@ -9,15 +9,15 @@ jobs:
...
+        uses: actions/checkout@v2
 
       - name: Lint and format CSS
         run: |
           npm ci
           npm run lint
 
       - name: Install PHP
-        uses: shivammathur/setup-php@2.15.0
+        uses: shivammathur/setup-php@2
```
akien-mga: Meh, why does it work for one and not the other?

Calinou: <https://github.com/actions/checkout>'s releases use a `v` prefix for Git tags, whereas https://github.com/shivammathur/setup-php doesn't.

I presume GitHub Actions can infer the version tag if it starts with `v`, but not if it doesn't start with `v`.

akien-mga: Yeah adding the `v` fixed it.

PR #379, opened 2021-12-19
Fix open source redirect on the homepage

fedorsykora:
The open source link in the features section of the main page (see picture below) previously pointed to `/features#collaborate` (same as the link next to it).
image
This commit changes it so that it points to `#get_involved`.
Alternatively, it would also make sense for it to point to [the contributing article on the wiki](https://docs.godotengine.org/en/stable/community/contributing/ways_to_contribute.html) or [the Github organization](https://github.com/godotengine).


Calinou (approved):

PR #381, opened 2021-12-24
More grammatically proper headline for the landing page

lmerriam:
"The game engine you waited for" is a bit awkward in English, it's all in past tense as though the addressee once was looking for an engine like this but is not necessarily seeking one right now. "The game engine **you've been waiting** for" is a more typical way of saying what I believe the original author meant. You can google the idiom and will find many examples acknowledging it, here is one of the first that pops up:  https://www.gymglish.com/en/gymglish/english-translation/the-moment-youve-been-waiting-for

### DIFF
--- a/themes/godotengine/layouts/home.htm
+++ b/themes/godotengine/layouts/home.htm
@@ -144,7 +144,7 @@ postPage = "{{ :slug }}"
   <div class="container flex eqsize responsive">
 
     <div class="padded">
-      <h2>The game engine you waited&nbsp;for.</h2>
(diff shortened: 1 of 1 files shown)


Calinou:
I think one reason we went for "The game engine you waited for" is that it broke better over two lines. This meant that it looked better visually, even if it's grammatically incorrect.

I'm not against this change, but we need to make sure the text still breaks over two lines in a good way (or force this somehow).

YuriSizov:
This is superseded by https://github.com/godotengine/godot-website/pull/409. Thanks for your contribution nonetheless!

PR #382, opened 2022-01-03
Fix images flickering when hovered in blog posts

### LINKED ISSUE #369: Image hover effect on blog triggers blinking when cursor is on the edge
Mageia 9 Linux, Firefox Nightly 94.0a1 from 2021-09-25.

**Issue description:**
We have a hover effect on images which highlights the fact that it's clickable, and on the blog, it leads to a weird bug as apparently there's a mismatch between hitboxes for normal and hover styles. I thought this was reported already but couldn't find the issue.

It's been like this for a while (since this hover effect was added).

The bug seems to happen only on the blog, but for any image (both banner picture and in-article images).

**Screenshots (greatly appreciated if it’s a style issue)**

https://user-images.githubusercontent.com/4701338/134922569-162f964a-9ac4-4387-8e2d-4e2ddfa1901d.mp4

Calinou:
This fix sets `filter` on all states, so that the browsers compute the selectable area in the same way regardless of the element's hover state.

Tested on Firefox 95 and Chromium 94.

This closes https://github.com/godotengine/godot-website/issues/369.

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -187,6 +187,13 @@ body a {
   transition: 0.1s filter;
 }
 
+body a,
+a.btn.flat {
+  /* Use `filter` on all states to prevent the clickable area from changing */
+  /* when an image is hovered (see GH-369). */
+  filter: brightness(100%);
+}
+
 body a:hover,
 a.btn.flat:hover {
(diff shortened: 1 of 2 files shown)


akien-mga (approved):

akien-mga:
Thanks!

PR #384, opened 2022-02-09
Update dates on page of upcoming events

byyiro:
When I found this page I thought it would be good to show updated dates even if we don't have any upcoming events yet as it feels less 'abandoned'.


Calinou (approved):
Thanks!

PR #386, opened 2022-02-19
Use current year in footer, update LICENSE to 2022

mhilbrunner:
- Use the current year in the footer automatically
- Update the license to 2022 in two places


Calinou (approved):
Thanks!

PR #394, opened 2022-05-17
Add information for Barcelona 2022 events.

JFonS:
Add information for Barcelona 2022 events. I didn't fully test this changes since I don't have the proper local setup. It would be good to test it in a local running instance before merging.


akien-mga (approved):

akien-mga:
Thanks!

Issue #33, opened 2018-01-31
Menubar at top of webpage is spam and should not be visible when reading while scrolled down on page


seronis:
Vertical screen space is a limited and precious commodity.   And it is being wrongfully used by a pointless menubar that refuses to stay out of sight at the top of the page where it belongs.

If the menubar was statically placed you can still access it at any time 'instantly'.  Pressing the Home key scrolls a page to the top.  Even on laptops that dont have a dedicated home key you still have the 'function' key that makes one of the other keys act as the home key which takes the exact same amount of time to type. It also takes less than a second to just drag the scroll bar to the top. On a phone (at least on my Android) a couple brisk swipes can scroll to the top of a webpage even if you were 30 'pages' of content down.

In other words, THERE IS NO EXCUSE OR REASON AT ALL to be wasting screen viewing space for something that is still accessible effectively 'instantly' on any platform.

timschellenberg:
That's a very outdated design guideline, especially for presentation-focused web pages. I would totally understand the complaint if godotengine.org was a page that you would power navigate for long periods at a time (like a message board or news site), but its entire focus is to look appealing and be convenient to navigate, it's not very concerned with conserving vertical space.

The big game screenshot taking up the majority of the vertical fold (eg. the portion of the site that is visible immediately upon loading) seems like a bigger concern to me, because on smaller resolutions there just isn't much of any information about what the site is actually about.

seronis:
No, the entire point is to be convenient to READ.   The bar being static is still equally convenient to navigate since its still instantly accessible no matter what.   But being static would mean its no longer IN THE WAY.   Stuff that is IN THE WAY doesnt look good.  It looks spammy.

timschellenberg:
I found it pretty convenient when I was reading the 3.0 release post and wanted to go to the download page without losing where I was in that long article, personally.
Of course it can't be perfect for everyone, but "spammy" or "in the way" is definitely not what I would call a 64 pixel high, often helpful menu bar.

Being instantly accessible isn't really a great argument because if you press the up or down arrow key or move the scroll wheel, the info that the bar is covering up becomes visible too, although the chances that you want to look exactly there are rather low because you'd have to scroll a LOT to cover the giant symbols on the front page completely.

Not to mention a lot of people aren't even aware that the home key does anything.


[...]

PR #406, opened 2022-07-14
Making the nav bar follow the scroll view

coppolaemilio:
Maybe I should have split this into two PRs, but I can split it if needed.

On this PR:
- Made it so that the nav bar follows your screen.
- Added `scroll-padding-top` to the `html` so that when you open an anchored link the nav is not covering the content
- Softened the nav bar shadow since it was too hard and small
- I added a small parallax effect on the home page so that it feels a bit less static
- Made the hero image a bit shorter because it was covering too much screen

https://user-images.githubusercontent.com/2206700/179080356-ecdb638c-d37c-4fb5-880e-af9846f7c1f2.mp4

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -311,7 +311,6 @@ h4 {
 
 header {
   /* Show on top of GodotCon banner (if any). */
-  position: relative;
   z-index: 2;
 
   width: 100vw;
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -51,6 +51,10 @@
   --teams-subteams-color: rgba(255, 255, 255, 0.4);
 }
 
+html {
+  scroll-padding-top: 64px; /* Navbar height */
+}
+
 .nav-logo.dark-logo {
   display: none;
 }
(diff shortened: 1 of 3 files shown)


Calinou:
I apologize, but fixed navigation bars are considered an antifeature nowadays for two reasons:

- It wastes screen real estate that is precious nowadays as 16:9 displays are the norm on desktop. As of 2022, browsers are still not that concerned about giving more vertical space for websites to display.
- It's considered [distracting](https://github.com/godotengine/godot-website/issues/33) by many.

This can be worked around with some JavaScript code to hide the navigation bar when scrolling down (and show it when scrolling back up), but this can interfere with browsers doing this to their address bars on mobile.

Most major websites like GitHub have stopped using fixed navigation bars around 2019-2020 (after using them for a few years).

Parallax scrolling also suffers from similar issues, most notably performance. It requires much more CPU cycles to update a parallax scrolling layer compared to a non-parallax one, which makes it hard for slow devices to keep updating at 60 Hz (let alone 120 Hz). This is especially noticeable on Firefox on Android (which is less optimized compared to Chrome on Android), but it can still be felt on low-end devices on Chrome.

Making the hero image shorter is a good idea though. I'm neutral about the navigation bar shadow change, but I think it's reasonable to try it out.

coppolaemilio:
I can do the scrollbar showing up only when scrolling up (like blender does) or just scrap it 👍🏻  I'll remove the parallax

PR #409, opened 2022-07-16
Update copy and media on the home page

bits-by-brandon:
## Homepage incremental upgrades
- Edits homepage cards copy to better reflect the questions of new users to the site
- Adds new card for Cross platform
- Adds new image for Cross platform
- Slight design tweaks to cards to make them look a little more appealing

draft


themes/godotengine/layouts/home.htm
```diff
@@ -143,10 +166,10 @@
   <div class="container flex eqsize responsive">
 
     <div class="padded">
-      <h2>The game engine you waited&nbsp;for.</h2>
+      <h2>The game engine you've been waiting for.</h2>
```
YuriSizov: See https://github.com/godotengine/godot-website/pull/381, there is a concern about how it would break into lines at different sizes.

themes/godotengine/layouts/home.htm
```diff
@@ -241,48 +270,57 @@ <h4>Gorgeous 3D</h4>
...
+            and render your 3D worlds and characters.
+          </p>
         </div>
       </div>
     </a>
   </div>
 
   <div class="flex eqsize responsive features-row">
-    <a href="/features#collaborate" data-barba-prevent>
+    <a href="https://docs.godotengine.org/en/stable/tutorials/export/exporting_basics.html#exporting-by-platform" data-barba-prevent>
```
YuriSizov: I would still link to a section on the Features page (that we need to create at the moment), so people don't leave the main website by accident. That section, in turn, can link to the documentation.

YuriSizov:
As mentioned in chat, I'm not sure if we are allowed to use platform logos, especially if we modify them to match our desired looks. There certainly was some discussion about that when we were updating platform logos in the editor, for the export templates. 😞

coppolaemilio:
It would be great to make the cards have the image followed by the text box but without any transparency.
Something like in the mockup I made here:

Screen Shot 2022-07-16 at 12 09 18

If the image is light the difference between the background transparency and the copy has not enough contrast to be accessible. 

We should also make the images more similar to a regular thumbnail (landscape) instead of a big square.

Let me know if you need some help with those changes and if it makes sense.

themes/godotengine/layouts/home.htm
bits-by-brandon: 🤔 IMO awkward line breaks are better than an slightly awkward phrasing as the very first text on the homepage, but I can remove for now

themes/godotengine/layouts/home.htm
bits-by-brandon: Didn't realize there was a section for deployments

YuriSizov: Ah, right, me neither, I thought we needed to create it still :D Even better, can totally be `#deploy`!

themes/godotengine/layouts/home.htm
bits-by-brandon: Okay fixed this so it no longer haves any widows

bits-by-brandon:
@Calinou  ready for merge!

coppolaemilio:
Since I can't add a commit to this PR I will leave here the changes I made to the home page to fix a visual bug.

There was a bug that caused some of the cards with shorter text to end before the siblings resulting in them having a different height.
I also moved the cards into a single element of a grid since it fits this purpose better than building each row and adding the margins individually as it was done before.

## File with changes
📄 [home.htm](https://github.com/godotengine/godot-website/files/9126861/home.txt)

## Before
Screen Shot 2022-07-17 at 02 18 08

## After 
image


I tested in mobile and desktop 👍 but more test is welcome

YuriSizov:
Looks good and responsive! I've squashed everything, fixed the mutliplatform link (it was to the editor and not to the deployment; while ok on desktop since they share the line, it wasn't correct for mobile), and also made the download button centered on lower resolutions. It bugged me when I was testing responsiveness

[Before](https://user-images.githubusercontent.com/11782833/179511435-8d9fe851-314a-4900-903c-a558e529efe3.png)
[After](https://user-images.githubusercontent.com/11782833/179511374-e8ad7bc5-c70c-4d73-ae58-3eeb2e933142.png)

YuriSizov (approved):

coppolaemilio (approved):

themes/godotengine/layouts/home.htm
```diff
@@ -143,21 +191,23 @@
   <div class="container flex eqsize responsive">
 
     <div class="padded">
-      <h2>The game engine you waited&nbsp;for.</h2>
+      <h2>The game engine you've been waiting&nbsp;for.</h2>
       <p>
-        Godot provides a huge set of common tools, so you can just focus on
-        making your game without reinventing the wheel.
+        Godot provides a huge amount powerful tools out of the box, so you can
```
akien-mga: Missing word.

Also feels a bit heavy handed to me but not fully opposed. This is **the** elevator pitch for Godot reused everywhere so we need to make sure we have a clear improvement here if we're changing it. I would suggest bringing it for review to other maintainers.

themes/godotengine/layouts/home.htm
```diff
@@ -143,21 +191,23 @@
...
+        just focus on making your game without reinventing the wheel.
```
akien-mga: I'd drop the "just".

themes/godotengine/layouts/home.htm
```diff
@@ -197,98 +247,108 @@ <h3 class="base-padding no-margin-bottom">News</h3>
...
           height="1"
           style="background: linear-gradient(90deg, #333747 11%, #2d3342 50%, #2d3342 68%, #272d3c 87%)"
           loading="lazy"
         >
-        <div class="dark base-padding">
+        <div class="body">
           <h4>Innovative design</h4>
-          <p>Big or small ideas adapt seamlessly to Godot's node-based architecture, making your life easier.</p>
+          <p>
+            Godot's Node and Scene systems give you both power and flexibility
```
akien-mga: Maybe "system" should be singular? Nodes and Scenes are two sides of the same coin, to me it's a single system.

themes/godotengine/layouts/home.htm
```diff
@@ -197,98 +247,108 @@ <h3 class="base-padding no-margin-bottom">News</h3>
...
+          style="background: linear-gradient(90deg, #252a35 46%, #252a35 53%, #202630 76%, #202630 89%)"
           loading="lazy"
         >
-        <div class="dark base-padding">
-          <h4>Gorgeous 3D</h4>
-          <p>Innovative 3D renderer design, which makes your art look great with minimal effort.</p>
+        <div class="body">
+          <h4>Use the right language for the job</h4>
+          <p>
+            Keep your code modular with an Object oriented API using Godot's own
```
akien-mga:
```suggestion
            Keep your code modular with an object-oriented API using Godot's own
```

themes/godotengine/layouts/home.htm
```diff
@@ -197,98 +247,108 @@ <h3 class="base-padding no-margin-bottom">News</h3>
...
           loading="lazy"
         >
-        <div class="dark base-padding">
-          <h4>Beautiful 2D</h4>
-          <p>Dedicated 2D engine that works in pixel coordinates, with plenty of built-in tools.</p>
+        <div class="body">
+          <h4>Dedicated 2D engine</h4>
+          <p>
+            Make crisp and performant 2D games with Godot's dedicated 2D
+            rendering engine with real 2D pixel coordinates and 2D nodes.
```
akien-mga, line 304: Sounds like a yogurt ad :P (*with real bits of fruit*) but it's good :+1:

themes/godotengine/layouts/home.htm
```diff
@@ -261,7 +261,7 @@ <h2>A different way to make games</h2>
         <div class="body">
           <h4>Innovative design</h4>
           <p>
-            Godot's Node and Scene systems give you both power and flexibility
+            Godot's Node and Scene system give you both power and flexibility
```
akien-mga: *gives*

akien-mga (approved):
Looks great!

akien-mga:
Thanks a lot!

PR #411, opened 2022-07-18
Cross-link the Teams page from the Governance page

YuriSizov:
This removes the note that teams will be published in the future (already published) and adds a link to the page. Swapped the first "Teams" for "Engine teams" so that the link stands out more.


akien-mga (approved):

akien-mga:
Thanks!

PR #413, opened 2022-07-18
Move the About link in the header to reduce significance

YuriSizov:
The "About" link is less important than other links in there, so this PR moves it deeper to the right (and mobile-only direct "Showcase" link is brought towards the top in turn). Assets is still right-most, though, because it catches the eye better like that, being on the border.

image


akien-mga (approved):

akien-mga:
Thanks!

PR #415, opened 2022-07-21
Update XR support section on Features page

dsnopek:


themes/godotengine/pages/features.htm
```diff
@@ -234,8 +234,11 @@ <h2>XR support</h2>
 	  <p>Godot makes cross-platform Augmented and Virtual Reality development easy.</p>
 
 	  <ul>
-		<li><strong>Plugin structure</strong> gives access to various SDKs such as the OpenVR, OpenXR and Oculus SDKs. Godot works with HTC Vive, Valve Index, Oculus Rift, Oculus Go, Oculus Quest, all Microsoft MR headsets, and many more.</li>
-		<li><strong>ARKit support</strong> is directly available in the iOS build.</li>
+		<li><strong>Works with many headsets</strong> including the Meta Quest, Valve Index, HTC Vive, Oculus Rift, all Microsoft MR headsets, and many more.</li>
+		<li><strong>Support for OpenXR</strong>, an emerging open standard, that most hardware vendors are moving to.</li>
+		<li><strong>Plugin structure</strong> gives access to various proprietary SDKs, such as OpenVR (SteamVR) and the legacy Oculus SDKs.</li>
+		<li><strong>WebXR</strong> allows developing AR and VR experiences accessible via a web browser.</li>
+		<li><strong>ARKit</strong> allows creating AR experiences on iOS.</li>
```
YuriSizov:
```suggestion
		<li>Ready for AR and VR experiences in a web browser with <strong>WebXR</strong> and AR experiences on iOS with <strong>ARKit</strong>.</li>
```

Maybe put those together to get read of a repeated pattern?

YuriSizov (approved):
I think texts are good, though I'd probably do the change I've mentioned in the comment.

themes/godotengine/pages/features.htm
dsnopek: Hm. They are two very different features, though. Maybe there's another way to keep them as separate bullet points but get rid of the repetition? I'll think about this a bit...

YuriSizov: I get that from the technical POV they are different, but I think for creating interest in users and for marketing the main point here is "We support XR on web and mobile too!". But if you want to keep them separate, but give them more flavor, go ahead. Let's wait for other to review this anyway

dsnopek: Well, my original hope was to give OpenXR and WebXR the most prominence, since they are open standards and will (eventually) be both supported in Godot core. But it just sort of turned out to flow better talking about supported headsets first then OpenXR, but I couldn't jump straight to WebXR, because the note about plugins for proprietary APIs only makes sense right after OpenXR. And now WebXR is down at the bottom with ARKit :-)

I've pushed some changes that make the wording on the WebXR and ARKit lines more unique. And it makes space for changing the last line to be:

```html
<li><strong>ARKit</strong> and <strong>ARCore</strong> allow creating AR apps for iOS and Android.</li>
```

... if the ARCore stuff ever gets cleaned up and finished :-)

Please let me know what you think!

themes/godotengine/pages/features.htm
```diff
@@ -234,8 +234,11 @@ <h2>XR support</h2>
 	  <p>Godot makes cross-platform Augmented and Virtual Reality development easy.</p>
 
 	  <ul>
-		<li><strong>Plugin structure</strong> gives access to various SDKs such as the OpenVR, OpenXR and Oculus SDKs. Godot works with HTC Vive, Valve Index, Oculus Rift, Oculus Go, Oculus Quest, all Microsoft MR headsets, and many more.</li>
-		<li><strong>ARKit support</strong> is directly available in the iOS build.</li>
+		<li><strong>Works with many headsets</strong> including the Meta Quest, Valve Index, HTC Vive, Oculus Rift, all Microsoft MR headsets, and many more.</li>
+		<li><strong>Support for OpenXR</strong>, an emerging open standard, that most hardware vendors are moving to.</li>
```
Calinou:
```suggestion
		<li><strong>Support for OpenXR</strong>, an emerging open standard that most hardware vendors are moving to.</li>
```

Calinou (approved):
Thanks!

PR #430, opened 2022-08-14
Fix City Game Studio trailer URL on showcase page

xsellier:
The video ID is TODO actually while the original trailer should be `hcRDBdvtuVQ`.

See the original trailer: https://www.youtube.com/watch?v=hcRDBdvtuVQ

Cheers


Calinou (approved):
Thanks!

PR #431, opened 2022-08-15
Update City Game Studio details

xsellier:
The game left early access in August 2021
It is also available on GOG since June 2022


YuriSizov (approved):

YuriSizov:
Thanks!

PR #433, opened 2022-09-14
improve search bar

alexfriesen:
This PR replaces the iframe search bar on the news page with a simple form which can be fully customised.

Everything should work the same as before.

### Preview:
image


Calinou:
Looks good to me, but make sure to use a dark color for the search bar background on the dark theme. It looks quite jarring otherwise.

alexfriesen:
Thanks for your reply.
I changed the styling a little.

### Preview:
---
image
image

YuriSizov:
So I noticed a regression because of this PR. It modified some styles that weren't actually valid before and didn't apply to anything (or at least didn't apply to some things), but now they do. See the giant spacing here:

chrome_2022-11-26_00-07-33

I'll look into it, so this is just noting.

PR #444, opened 2022-10-10
Article Page improvements

alexfriesen:
This will add round borders to images, videos, iframes and code.

Also headlines will have a word-break.

### DIFF
--- a/themes/godotengine/layouts/article.htm
+++ b/themes/godotengine/layouts/article.htm
@@ -30,6 +30,7 @@ categoryPage = "article"
   pre {
     background: var(--codeblock-background-color);
     color: var(--codeblock-color);
+    border-radius: 0.5rem;
   }
 
   pre code {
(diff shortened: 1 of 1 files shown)


YuriSizov:
It seems to me that code blocks already have rounded corners, just with a smaller factor. I don't mind either way, but for the rest of it, I'm not sure it improves things, especially with a big value for the curvature. Our media files aren't designed for any specific style, but with rounded corners a lot of them start to look cropped and not too appealing.

It can also get very weird sometimes (notice also a visual issue that appears from the roundedness):

image

> Also headlines will have a word-break.

This may be worth it, but can you point to a page that showcases the issue? Or is it something that becomes really apparent on smaller screen devices, on mobile?

coppolaemilio:
I agree with Yuri that rounder corners on images inside the post don't make a lot of sense. I think since this PR is a bit old and there hasn't been a follow up we could close it.

YuriSizov:
Thanks for the contribution nonetheless! I'm still curious if there is an article that highlights an issue with the titles that requires your fix (or some other fix, perhaps).

alexfriesen:
Sorry for the late reply, I was quite busy.
Closing this is fine. I should have broken up those changes in separate PR's and explain better what those changes supposed to do.

PR #447, opened 2022-10-19
adjusts article image sizing to be the same width as content

bits-by-brandon:
# Summary
Slight adjustments to the hero image on articles to allow them to flow to the full width of the content

Screen Shot 2022-10-18 at 8 26 31 PM
Screen Shot 2022-10-18 at 8 26 45 PM
Screen Shot 2022-10-18 at 8 26 54 PM

### DIFF
--- a/themes/godotengine/layouts/article.htm
+++ b/themes/godotengine/layouts/article.htm
@@ -65,10 +65,6 @@ categoryPage = "article"
     margin-bottom: 16px;
   }
 
-  article img:first-child, article video:first-child {
-    max-width: 85%;
-  }
-
   article h1 {
     margin-top: 64px;
   }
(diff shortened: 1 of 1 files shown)


coppolaemilio:
I like how it looks but on desktop I would still go full width instead of left justified. Not sure if it would make the images massively big, but I think it will look better. Not sure what you think 😅

bits-by-brandon:
Wow missed that. Yeah i'll make it full width

bits-by-brandon:
@coppolaemilio Ready for another pass with the full width
Screen Shot 2022-10-31 at 6 24 15 PM

bits-by-brandon:
@coppolaemilio ready for rreview!

coppolaemilio:
Sorry about the delay reviewing this, I'll take a look now!

PR #449, opened 2022-10-31
Removing the name attribute from the OG image meta.

coppolaemilio:
The og image is not being indexed by sites like Reddit. By removing the `name="twitter:image"` it should get the preview image properly.

I'm not sure if we will also need to specify the `<meta property="og:image:type" content="image/jpeg" />` one, but I think we should try with this one first and see if it works. There isn't a way to test this locally, so we need to give it a go as-is.


coppolaemilio:
It seems like this is a Reddit bug, not an issue from our site.

PR #450, opened 2022-11-04
Showcase: Primal Light is releasing on Switch

akien-mga:
It will be released on 2022.11.24.

I was notified of it by porting/publisher [No Gravity Games](https://nogravitygames.com/) and there website indeed confirms the upcoming release.

### DIFF
--- a/themes/godotengine/pages/showcase.htm
+++ b/themes/godotengine/pages/showcase.htm
@@ -127,7 +127,7 @@ is_hidden = 0
       partial "showcase/list-item"
       title="Primal Light"
       author="Fat Gem"
-      platforms="windows, macos, linux"
+      platforms="windows, macos, linux, switch"
       slug="primal-light"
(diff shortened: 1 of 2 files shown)


Calinou (changes requested):
The `primal-light.html` details page should also be updated to have the `switch` icon. I know it's not very DRY but it was the easiest way to implement the showcase :slightly_smiling_face:

akien-mga:
Thanks for the review, it should be correct now I believe. Note that I didn't test the changes.

Calinou (approved):
Good to merge ~~**after** 2022-11-24 (don't merge before).~~

akien-mga:
I think it's fine to merge even before, it shouldn't make much difference.

There's already a store page so users who notice this before Nov 24 could wishlist it: https://www.nintendo.com/store/products/primal-light-switch/

PR #455, opened 2022-11-16
Make styling of news tags more consistent

YuriSizov:
I've always found the styling that we have for tags/categories to be inconsistent with the rest of our design. So here's an attempt to make it a bit more fitting:

**Article list**
[Before](https://user-images.githubusercontent.com/11782833/202274830-dc151ed4-24de-41a0-96d9-af489dfcc78b.png) | [After](https://user-images.githubusercontent.com/11782833/202274846-1e962201-c798-403f-93e8-52d75548da5a.png)

**Individual article**
[Before](https://user-images.githubusercontent.com/11782833/202274861-5919924a-2738-48c1-be91-c12dfd34e848.png) | [After](https://user-images.githubusercontent.com/11782833/202274875-ae7de2d2-0a72-4a76-8c86-7d2171073948.png)

This makes use of the Monserrat font that we use for titles and such. Initially I wanted to go with the default font and all-caps tags, but this is a good alternative if we don't like using all-caps for usability reasons. This also removes a lot of accent on inactive items on the article list page. Accent is also slightly reduced on the individual article page, so it's not being too distracting. Extra box shadow, using pre-existing styling, makes the buttons pop a bit more, which I find beneficial (e.g. it breaks up the article page better).

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -776,10 +776,20 @@ p.small {
 
 .tag {
   display: inline-block;
-  border-radius: 1000px;
   background-color: var(--card-background-color);
+  border-radius: 1000px;
+  box-shadow: var(--base-shadow);
+  filter: saturate(0.35);
+  font-family: "Montserrat", sans-serif;
+  font-size: 16px;
   padding: 0.25rem 0.75rem;
   margin: 0.1rem;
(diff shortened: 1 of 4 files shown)


Calinou (approved):

coppolaemilio:
Thank you! :)

PR #460, opened 2022-11-21
Update `security.txt` expiration date

Calinou:
The date is set to be less than a year into the future as recommended per the specification.


akien-mga (approved):
Looks good.

How can we make sure that we're notified of the need to update it again in a year?

Calinou:
> How can we make sure that we're notified of the need to update it again in a year?

I'm not sure if there's a GitHub action for this or similar. Of course, I can a reminder in my calendar :slightly_smiling_face:

coppolaemilio:
Thanks!

Issue #64, opened 2018-09-19
Center the images uses as article thumbnails


groud:
In the [article list](https://godotengine.org/news), each thumbnail is basically a square subset of the top banner shown when you open the article. However, the square is taken from the left of the banner, which, in most case does not fit really well. Indeed, the most representative part of the banner is usually the center.

I suggest we use the center of the thumbnails instead of its left part as a thumbnail.

Calinou:
It should be as easy as adding `background-position: center` to all article thumbnails (in the same class where `background-size: cover` is defined, I guess).

```css
background-position: center;
```

**Preview:**

![image](https://user-images.githubusercontent.com/180032/45753257-74d2a900-bc18-11e8-8a70-bbc7fc92584e.png)

MHillier98:
This would be a good change for some banners, but for others, it would appear this has been used as a feature:

**Example 1:**
![image](https://user-images.githubusercontent.com/26564949/54475432-67a01b00-4845-11e9-9fea-6abf629d9638.png)

Changes to:
![image](https://user-images.githubusercontent.com/26564949/54475423-5951ff00-4845-11e9-8616-6f05a06221ae.png)

_However,_

**Example 2:**
![image](https://user-images.githubusercontent.com/26564949/54475441-82728f80-4845-11e9-898a-640377806879.png)

Changes to:
![image](https://user-images.githubusercontent.com/26564949/54475450-8c948e00-4845-11e9-9940-cf6f3c3ab01b.png)


This would have to be a change to how it's being used in function, or some edits would have to be made. Perhaps a better solution would be to add a preview image for blog posts? But that does seem like a fair bit of work to backfill images, and the added storage space for the images.

akien-mga:
It wasn't really used as a feature, though you're right it turned out well in those two examples :)

If we change to center, I'll just keep this in mind when picking up banner images to ensure that the cropped version would look decent.

MHillier98:
Another potential solution could be to make the images not take up the full height, allowing them to maintain their original size? However I do assume that it was designed to be a square originally:

![image](https://user-images.githubusercontent.com/26564949/54483941-88518a80-48b0-11e9-9f0b-6455e3df157e.png)

The CSS for the images:
```
element {
	background: url('https://godotengine.org/storage/app/uploads/public/5c7/67d/8c6/thumb_667_400x1_0_0_auto.png');
	background-position: center;
	background-repeat: no-repeat;
	background-size: contain;
}
```

[...]

PR #468, opened 2022-11-25
Updating news to display articles on a grid

coppolaemilio:
A small redesign of the news page to show the thumbnails without cropping them, have more articles per row.
Images are centered and 16/9 aspect ratio, so if they get cropped, the most important elements will be displayed.
It starts with a single column in mobile, and then it goes to two columns on 768px, and three after 1024px.
Avatars get pulled from the CMS and are called by using the provided `{{ post.user.getAvatarThumb() }}`

image

* The articles on the screenshot are placeholders I made locally

This is how it looks like in the ~~eye strain~~ light theme:

image

I considered adding tags and other information but ultimately decided to leave it as clean as possible. 

For reference, this is how it looks right now: 

image

There are still other areas of the News page I would like to update, so this is the first in a series of updates.


Calinou:
Looks good, but the author/date text isn't very readable on the dark theme. Could you make it a bit brighter when using the dark theme? Once this is done, this should be good to merge :slightly_smiling_face:

coppolaemilio:
> Looks good, but the author/date text isn't very readable on the dark theme. Could you make it a bit brighter when using the dark theme? Once this is done, this should be good to merge 🙂

Thanks! I tried using a color that was already a variable used on other text, but it is true that it doesn't have a lot of contrast with the background. I'll find another value and update it 👍

coppolaemilio:
This is how the new placeholder avatar looks like and also more contrast on the info area:

image

@YuriSizov @Calinou should be ready to merge/publish whenever you are ready

YuriSizov (approved):
I've changed one small thing: avatars now load at 40 pixel, but display at 25 still. This reduces compression on this small image by a lot. Otherwise, I think this is good to go!

Note, that we would want to check the avatars upon going live, and maybe upload a placeholder for those that may be considered personal. We can keep the ones that match other public avatars of people (on GH, Twitter), or something abstract enough.

YuriSizov:
Okay, I made it so Gravatar is never used, so our authors won't accidentally get assigned an image they'd rather not have. And we can fallback without it, so we're using an SVG now!

YuriSizov:
Just one tiny fix that annoyed me about this page. Focused style for the tags is now correct:

image

It used to look like [this](https://user-images.githubusercontent.com/11782833/204104002-cbe1ff0b-e836-4701-b425-f38f9bcdcaef.png). We may want to make sure all focus styles look appropriate for their corresponding elements, but this was was really out there for some reason. And since this page will get more attention when we launch this, I decided to quickly fix it.

-----

The PR is as ready as it gets, I'm just taking the opportunity to tweak it even more before it gets merged. 🙃

coppolaemilio:
This PR closes https://github.com/godotengine/godot-website/issues/64

YuriSizov:
Thanks a bunch!

PR #473, opened 2022-11-28
Update address for matrix room

Welteam:
I want to make clear that the current link is not broken as it redirects to the correct room.


coppolaemilio:
Thank you @Welteam

PR #475, opened 2022-12-06
Home refresh: Hero, news and features sections updated.

coppolaemilio:
This is the first PR in a series of PRs aimed to update the home page.

## Hero section
The image will get selected randomly each time you visit the home page.
Credits and link to the games is found on the bottom right of the image.
The nav is now transparent (only on this page) but it will have to get reworked in a future PR.

Screen Shot

## News section
The news section now displays each article using the new card design recently added to the news page.

Screen Shot

## Features section

I modified the features section to be 3 columns instead of 2.
This makes it look more in sync with the next section and since the cards are smaller now, they are easier to read. Before, an image could take the entire screen and it was hard to read the text while looking at the image above.

Screen Shot 2022-12-06 at 03 11 08

Those are all the changes for now, but to continue with more, I need to do more theme updates first. Those will come in a future PR.


coppolaemilio:
I'm open to discuss the functionality of the hero image. I think random works best so we don't have to decide which to highlight, but I couldn't find a good solution that would work and also keep it exciting.

YuriSizov:
Amazing work, thanks!

Edit: Oops, I forgot to approve! 😄 Thank god we have no checks.

PR #476, opened 2022-12-07
Add a custom utility plugin

YuriSizov:
As a first step to revive the internationalization efforts (https://github.com/godotengine/godot-website/pull/178) this PR adds a custom CMS plugin. I'm adding an empty plugin first to test that it will correctly work in production and in dockers for other contributors.

This plugin comes with a CLI tool that simply responds to the caller:
```
php artisan godotengine:ping
```

It should have no visible effect on anything, but after deploying it we need to check if the plugin is activated correctly or if it needs a nudge. Running the aforementioned command should be enough. If it doesn't work or isn't recognized out of the gate, we should try `php artisan winter:up` first. May create a momentary downtime, but shouldn't break anything. Maybe it's worth to make a backup of the directory...

Locally we still use October, so you should call `php artisan october:up` as needed. But first of all you would need to recreate the docker container following the normal instructions in README.


YuriSizov:
Just a short note that CLI API doesn't seem to work for a mysterious reason. I've debugged it for 3 days, including going through PHP sources, but the ultimate result is that I don't know why it doesn't work, and we can't adjust the CMS code to a different method that works, because that would lock us out of updates.

I should be able to make the same functionality to be available through the CMS' control panel instead. Just need a bit more time to figure it out.

PR #477, opened 2022-12-07
Adding dome keeper

coppolaemilio:
In this PR I added Dome Keeper to the list of showcase games, and it is also a featured game on the home page.
Since some of the screenshots on the showcase were the same as the ones used as hero images, I removed the repeated ones. The entire showcase section + hero should be a collection in the CMS instead of data manually added on PRs, but that will be in a future update.
I also made it so that the scroll on mobile doesn't trigger the parallax effect.


YuriSizov:
> I also made it so that the scroll on mobile doesn't trigger the parallax effect.

I've separated this change in a dedicated commit and also added following changes:

* The parallax effect is disabled for users who opted to have reduced motion.
* The background offset of the parallax image is updated on window resize, so it looks correctly when you go between desktop, tablet, and mobile modes.
* The frosted glass effect should now work in Safari on both iOS and macOS.
* In mobile/tablet mode the main navigation is now a little better spaced.
* Made an attempt to prevent our donate "heart" character from turning into an emoji on mobile devices to preserve our style.

Still unknown what causes the reported flicker for some users. Hopefully it's because of the parallax and is now fixed, but I can't repro it myself on either iPhone or iPad.

coppolaemilio:
LGTM 👍 no flickering on my end

YuriSizov (approved):

PR #480, opened 2022-12-10
Buttons fully functional, but for more scalability and readability, database changes should be made for this issue.

Alex-Reynolds:
The buttons are fully functional. This feature was possible with the current database, but the solution is a little bit messy. For better code maintainability, this feature should be added with a reformatting of the database, so that the posts/articles hold two extra properties. One would be a reference to the next article and the other property, would be a reference to the previous article. 

---disclaimers
This solution relies on the assumption that the Asset Library's collection of posts are ascending order by id. 
If that is not the case, this solution will most likely require adding those two properties to the posts/articles.
This solution will need to be updated if somehow the number of posts on Godot became greater than 10,000


Calinou:
Superseded by https://github.com/godotengine/godot-website/pull/481.

Issue #419, opened 2022-08-01
Add links to next and previous news articles


JohnVeness:
When reading individual Godot news articles (e.g. those linked to from https://godotengine.org/news/), it would be nice if there were links to the next and previous articles.

So for example, in the [Godot and consoles, all you need to know](https://godotengine.org/article/godot-consoles-all-you-need-know) article, there would be links at the bottom to [Release candidate: Godot 3.4.5 RC 1](https://godotengine.org/article/release-candidate-godot-3-4-5-rc-1) and [Dev snapshot: Godot 4.0 alpha 12](https://godotengine.org/article/dev-snapshot-godot-4-0-alpha-12).

This would allow someone to read articles in order, without having to keep pressing Back and then picking the next article. This would be most useful to someone who wants to catch up with a lot of Godot news (e.g. someone new to the engine).

Hopefully this is something the CMS can do for you - I'm not suggesting you manually add all the links!

See for example the site https://retrostuff.org/2021/09/27/sega-pico-flash-cart-adapter-v2/ which does this.

Thanks.

Alex-Reynolds:
I am part of a team for a university class, we would like to attempt to add this feature.

YuriSizov:
@Alex-Reynolds Well, nobody else is working on it yet. But we will probably work on an update to the page overall, which might incorporate something like that.

Feature-wise, it's probably more useful to display relevant news (using the tagging/category system, perhaps) than just next and previous news.

Alex-Reynolds:
We don't require our PR to be merged. Hopefully our effort can help with an update though!

coppolaemilio:
I think adding the next and previous links could be a nice first step, and it can later on be improved to be the recommendation based on topics that @YuriSizov describes.
Good luck with the PR!

nsundriyal62:
i don't think it still resolve, can i work on this issue ?

Calinou:
Yes :slightly_smiling_face: 

The website now uses a static site generator (Jekyll), instead of a CMS (Winter). This will change *how* you should implement the feature, but it's still feasible.

nsundriyal62:
ok

i have a plenty beginner knowedge of jekyll and still learing it 

so i can resolve it


thanks man

PR #481, opened 2022-12-10
Added next and previous buttons for articles #419

ryantullis:
Buttons fully functional, but for more scalability and readability, database changes should be made for this issue

The buttons are fully functional. This feature was possible with the current database, but the solution is a little bit messy. For better code maintainability, this feature should be added with a reformatting of the database, so that the posts/articles hold two extra properties. One would be a reference to the next article and the other property, would be a reference to the previous article.

---disclaimers
This solution relies on the assumption that the Asset Library's collection of posts are ascending order by id.
If that is not the case, this solution will most likely require adding those two properties to the posts/articles.
This solution will need to be updated if somehow the number of posts on Godot became greater than 10,000

### DIFF
--- a/themes/godotengine/layouts/news.htm
+++ b/themes/godotengine/layouts/news.htm
@@ -1,3 +1,4 @@
+;;
 description = "News layout"
 ==
 {##}
--- a/themes/godotengine/pages/news.htm
+++ b/themes/godotengine/pages/news.htm
@@ -7,7 +7,7 @@ is_hidden = 0
 [blogPosts]
 pageNumber = "{{ :page }}"
 categoryFilter = "{{ :slug }}"
-postsPerPage = 24
+postsPerPage = 3
 noPostsMessage = "No posts found"
 sortOrder = "published_at desc"
 categoryPage = "article"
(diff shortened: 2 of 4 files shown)


themes/godotengine/layouts/news.htm
```diff
@@ -1,3 +1,4 @@
+;;
```
Calinou, line 1: Is this line useful for anything specific?

themes/godotengine/pages/news.htm
```diff
@@ -7,7 +7,7 @@
 [blogPosts]
 pageNumber = "{{ :page }}"
 categoryFilter = "{{ :slug }}"
-postsPerPage = 24
+postsPerPage = 3
```
Calinou, line 10: Was this number reduced for testing purposes? Does it affect how the list of news articles displays on <https://godotengine.org/news>?

Alex-Reynolds: Yes, That was added for debugging purposes. Should be changed back to 24

YuriSizov:
Needs to be redone for the new static website.

PR #482, opened 2022-12-11
Implement GodotEngine.I18n plugin

YuriSizov:
This is a revival of https://github.com/godotengine/godot-website/pull/178, if we decide to stick with Winter/October for now.

Compared to the old implementation this PR implements all the necessary management utilities via the backend control panel. We have an issue with PHP CLI on the production server, so it wasn't viable to go that way. I can restore the CLI commands for some dev environment convenience later, if needed.

The code has been refactored, split up and cleaned up. It also has been extensively documented, more so than before. As making a plugin relies on several libraries, which you may not be able to debug or inspect locally, I hope code comments would at least explain things a bit.

This PR doesn't include any user-facing UI for switching languages.

## Prerequisites

Our new plugin relies, as before, on `Rainlab.Translate`, it even injects itself into some logic to keep both plugins in sync. Therefore you can refer to its sources:

* https://github.com/rainlab/translate-plugin/tree/1.x

From Winter itself these two repositories can be of use:

* https://github.com/wintercms/storm/tree/1.0
* https://github.com/wintercms/winter/tree/1.0

And for interacting with PO files we're using a PHP binding for `gettext`:

* https://github.com/php-gettext/Gettext/

## Docker update

When testing locally you would need to recreate your docker container. Or you can run these commands:

```shell

    apt-get install gettext
    php artisan plugin:install "rainlab.translate"
    composer update -d ./plugins/godotengine/i18n
    php artisan october:up
```

## Content translation

This process should be the same as described in the original PR, and it comes with all the same considerations and limitations (read the part about markup injection and vulnerability concerns).

This PR only includes one line as an example:

```
<h1>{{ TR('The game engine you\'ve been waiting&nbsp;for.') }}</h1>
```

## Translation management

Same as before, PO files are tracked in the repo. The CMS itself uses YAML files which we need to keep in sync, but we don't need them in the repository.

As a first step you open the backend, navigate to **Settings > Translate > Godot I18n** and press **Setup** and **Refresh Locales**. This should make sure your local installation (as well as our production installation, if we add new languages) is up to date. Note that **Setup** is only available if all localization files are missing, so you are unlikely to see it. Still, it exists just in case.

You can then press **Extract Messages** to extract all translatable strings into PO, merging them in the process with the existing files, so we don't lose existing translations. You need to press this button after adding new translatable strings, modifying existing strings, or removing them from pages, layouts, partials, and other templates.

After PO files in the repository have been updated, you can press **Update Translation** to regenerate YAML files and refresh CMS database. You need to perform this step on the running instance to see the changes, so you can do it locally to test, and you need to do it on the production server after the changes have been pulled there as well.

That's pretty much it. New languages can be added via the `Rainlab.Translate` menus, navigate to **Settings > Translate > Manage Languages** and add one. This will generate the necessary files. You could also create them on disk yourself, and then press "Refresh Locales", but that would be peculiar.

chrome_2022-12-11_23-28-33


YuriSizov:
It should be good for local testing, if anyone is brave enough! I need to double-check if our docker setup still works, as there were some issues with the order of operations on CI which may be repeatable with the docker.

YuriSizov:
Docker indeed had the same issue. I've worked around it in a slightly ugly way, but I couldn't find anything nicer due to the order of operation and the need to mount folders.

Note to self: Before pulling this PR to the live instance I need to manually install the Rainlab.Translate plugin.

YuriSizov:
Changed most of the templates to use the actual language in the HTML attributes. I decided to not change the maintenance template because I wasn't sure the application and its translations would be available there. Also not sure we use it.

Additionally, injected link alternative tags using Rainlab's provided widget. Those don't play well with barba, because it doesn't update them on navigation. But this shouldn't cause issues for scrapper bots, I don't think.

chrome_2022-12-12_16-24-34

coppolaemilio (approved):

YuriSizov:
Deploy note: I need to mimic the same functionality as the one available from Winter/October CLI, so I can install new plugins and correctly migrate the database. It shouldn't be hard, I'll make another PR where I'll implement it as a backend screen from the Utility plugin. It would need to be deployed first, then the translate plugin needs to be installed, then this can be deployed.

It's a lot of activity, but there are some virtualization-related limitations with our hosting that we can't resolve internally. So we make do with what we have.

PR #491, opened 2022-12-14
Home page, nav, and footer translated to Spanish

coppolaemilio:
This PR adds the keys for the home page, the navigation bar and the footer sections.
I added an initial Spanish translation of the site. I had to figure out some of the marketing terms that might change in the future. I chose to keep _Asset_ as the English word because there isn't anything technical in Spanish that describes this concept in particular as far as I know.

This PR also includes a fix to the hero images not working on `/<lang>` nor `/<lang>/`.

image

### DIFF
--- a/themes/godotengine/layouts/home.htm
+++ b/themes/godotengine/layouts/home.htm
@@ -475,8 +475,7 @@ Make sure to update this gradient when the home background image is changed.
   <div class="container sm-full">
     <img id="sfc_graphic" src="{{ 'assets/home/sfc.svg' | theme }}" alt="Software Freedom Conservancy logo" width="1" height="1" class="img-auto-size"  loading="lazy">
     <h3 class="text-center">{{ TR('Donate') }}</h3>
-    <p class
-    ="small auto-margin">
+    <p class="small auto-margin">
       {{ TR("You don't need to be an engine developer to help Godot. Consider donating to speed up development and make Godot&nbsp Engine even more awesome!") }}
     </p>
 
--- a/themes/godotengine/i18n/po/messages.es.po
+++ b/themes/godotengine/i18n/po/messages.es.po
@@ -7,6 +7,10 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Godot Engine official website\n"
 "Report-Msgid-Bugs-To: https://github.com/godotengine/godot-website\n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
+"Last-Translator: \n"
+"Language-Team: \n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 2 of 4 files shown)


themes/godotengine/layouts/home.htm
```diff
@@ -329,10 +329,9 @@ <h2>A different way to make games</h2>
           loading="lazy"
         >
         <div class="body">
-          <h4>Innovative design</h4>
+          <h4>{{ TR('Innovative design') }}</h4>
           <p>
-            Godot's Node and Scene system gives you both power and flexibility
-            to create anything.
+            {{ TR('Godot’s Node and Scene system gives you both power and flexibility to create anything.') }}
```
akien-mga: I'm not sure it's a good idea to use the typographical quote here. It's pretty difficult to type for most normal users and thus won't be easy to reproduce in translations, nor in new strings we write.

coppolaemilio: I can replace them with regular ones, I like them better visually but not a big deal if we use the easier ones to type 😅

YuriSizov: There is something to be said about using proper punctuation for some languages. Like we use «» when we want to be "proper" and fancy with quotation marks, but I think it requires complete commitment, or it will end up being inconsistent.

akien-mga: In my experience, this character is always used inconsistently because not everyone cares / knows how to type it. It's usually less work to remove the few added by mistake (typically from people who copy pasted content from MS Word which may auto-insert them in some locales) than the other way around.

coppolaemilio: All the instances of `’` should be replaced by `’` on the stuff I changed/translated 👍

themes/godotengine/i18n/po/messages.po
```diff
@@ -257,3 +229,31 @@ msgstr ""
 #: godotengine/partials/footer.htm:9
 msgid "contributors"
 msgstr ""
+
+#: godotengine/layouts/home.htm:324
```
akien-mga: That diff is interesting as it shows that the PO generation tool doesn't produce a deterministic sorted list of strings, it adds new ones at the end.

coppolaemilio: Not sure how they are generated, I just press the buttons on the CMS 😅 but maybe generating them alphabetically might be a solution

themes/godotengine/layouts/home.htm
```diff
@@ -475,8 +475,9 @@ <h4>{{ TR('Report') }}</h4>
   <div class="container sm-full">
     <img id="sfc_graphic" src="{{ 'assets/home/sfc.svg' | theme }}" alt="Software Freedom Conservancy logo" width="1" height="1" class="img-auto-size"  loading="lazy">
     <h3 class="text-center">{{ TR('Donate') }}</h3>
-    <p class="small auto-margin">
-      {{ TR('You don’t need to be an engine developer to help Godot. Consider donating to speed up development and make Godot&nbsp Engine even more awesome!') }}
+    <p class
+    ="small auto-margin">
```
YuriSizov: What happened here?

coppolaemilio: no idea, probably a rogue enter I pressed by mistake 🤔

themes/godotengine/i18n/po/messages.po
YuriSizov: Will be fixed by https://github.com/godotengine/godot-website/pull/494.

YuriSizov (approved):
This is going to be amazing!

YuriSizov:
Needs to be redone for the new static website. Translations can be salvaged, but we don't have the framework for them yet.

PR #485, opened 2022-12-13
Improve performance of the parallax effect

bits-by-brandon:
Fixes https://github.com/godotengine/godot-website/issues/483

https://user-images.githubusercontent.com/9118169/207452374-eb7a0137-6371-4abf-9d44-4df5d5c94353.mov

PR #495, opened 2022-12-17
Fix the parallax effect animation frame binding

YuriSizov:
Seems like https://github.com/godotengine/godot-website/pull/485 contained a mistake. `requestAnimationFrame` takes a function as an argument, and instead the function was just called directly with its result being passed to `requestAnimationFrame` with no useful effect. It also spammed the console.

Not sure if the desired performance was achieved in the end, as it didn't actually run on animation frames. But it did run on the handled events themselves, so visually there was no difference for most users. I did notice a hitch on the first scroll though, a few times. Might be related.

Either way, this should be better now. It still works, doesn't spew errors, and hopefully removes any visual lagginess from less powered/lower refresh rate devices.

### DIFF
--- a/themes/godotengine/partials/footer.htm
+++ b/themes/godotengine/partials/footer.htm
@@ -189,20 +189,20 @@ description = "footer partial"
                 const parallaxMinWidth = 768;
                 const parallaxSpeed = 0.4;
 
-                const parallaxTick = (offsetY) => {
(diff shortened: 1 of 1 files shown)


coppolaemilio (approved):

coppolaemilio:
Thank you and sorry for the delay on this one! 😅

PR #496, opened 2022-12-17
Add a locale picker widget to the footer

YuriSizov:
Only available for the authorized users for now, so we can test it without making translations completely public (they'll still work if users change the URL directly).

This PR also makes barba.js update some meta information on navigation, such as the lang attribute of `<html>` and alternate links in the `<head>`. This should make transitions more complete, as well as enable the locale picker to work with barba.

Looks like this:
[Desktop](https://user-images.githubusercontent.com/11782833/208256426-7d981327-7db6-44bf-8000-e01df7cdc38d.png) | [Mobile](https://user-images.githubusercontent.com/11782833/208256438-d49ec2ec-4df8-407b-be31-b33f721143e8.png)

### DIFF
--- a/themes/godotengine/assets/css/main.css
+++ b/themes/godotengine/assets/css/main.css
@@ -43,6 +43,7 @@
   --button-background-color: #ffffff;
   --footer-color: rgb(175, 180, 220);
   --footer-background-color: #414d69;
+  --locale-background-color: #313849;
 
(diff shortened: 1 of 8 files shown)


YuriSizov:
This is irrelevant now, but can still be salvaged layout and design-wise when we get back to i18n.

PR #499, opened 2022-12-31
Complete website rewrite

coppolaemilio:
## This PR
This PR replaces the current CMS website with a static version of it built using Jekyll.

- Everything should be working 1 to 1 with the same urls except for the `/news` page (not articles) which was renamed as `/blog` (old links will redirect to new ones)
- The showcase entries on the home page now redirect to their showcase page in our site so that the user can choose which store front or page to visit next
- The rss.xml file now only shows the last 24 posts (same as the first page of the `/blog` page). This was done to decrease the overall weight of the file which was including everything since the first post
- I  changed the first link from the nav bar to go to the showcase page since the features one is still pretty outdated

## Test appreciated!
I tested as much as I could and compared with the current live site, but I might have missed something. If you notice please let me know or send a commit to this PR.

## To do in future PRs:
Many files inside the `/storage` directory are not properly compressed or are not being used on our site, so in the future we should start removing them. Hosting some of the files externally could also help reduce the size of the repo and speed up the build times. Since Jekyll copies the files to the `_site` directory, this `storage` situation slows down the process of building a new version each time.


YuriSizov (approved):
I think this can be merged and then tested more and iterated more. We've practically stopped working on the current version of the website, so I see no harm switching the main branch to this version. For the current production use we can start pulling from the `legacy-cms` branch if we need to. Also can make PRs against it.

When this is merged, remove `feature-static-jekyll` from this repo, which was used for temporary testing of the CI.

YuriSizov:
The current deployment procedure is as follows:

- Approve the PR and merge it into `master`.
- Wait for the CI to finish building and pushing to the `published` branch.
- Deploy the `published` branch to production the same way as we did before.

In future we can look into automating it, but the important thing is that the `master` branch is for development and sources, and `published` contains the static version itself, squashed into one single commit. That's what should be fetched to the live server.

coppolaemilio:
Let's do this 🚀

PR #501, opened 2023-01-10
Update showcase page with new layout and entries

coppolaemilio:
The second update to the showcase page. Next update will bring filtering of entries, but this time I wanted to start using the game's capsules and remove the text over the screenshots. I believe this makes is more similar to how they look on store-fronts like Steam so users will be more comfortable browsing the titles instead of reading the names on top of a generic screenshot. 

I also replaced the video embeds at the top with links to YouTube so that we don't load them if the user is not going to be playing them. This should speed up the loading time and avoid having 3 video players at the start of the page (no one is going to watch the video in the thumbnail size, it's better that they see it on the proper app).

There are two new games: Brotato and Endoparasitic.

In this PR I also separate the unreleased games to its new section.

Screenshot:
Screen Shot 2023-01-10 at 01 00 44


YuriSizov:
I'll give it a better look after we start deploying the new website, but from just looking at the screenshot — I'd split the platforms and the author, aligned them to the opposite ends of the cover picture.

Edit: Also "Hive Time" is hard to read. Might I suggest dimming the background under the text?

coppolaemilio:
> I'll give it a better look after we start deploying the new website, but from just looking at the screenshot — I'd split the platforms and the author, aligned them to the opposite ends of the cover picture.

That was my first thought, but it gets confusing when you have them on a grid since you don't really know where it belongs:

image


The Hive Time and other like those have text shadow, I could try making the entire image a bit darker or add more text shadow

coppolaemilio:
I also believe that splitting the Apps/tools to another section will make it better, so I'll make those edits later on this or a future PR

YuriSizov:
> That was my first thought, but it gets confusing when you have them on a grid since you don't really know where it belongs:

Can inset them a bit.

> The Hive Time and other like those have text shadow, I could try making the entire image a bit darker or add more text shadow

Making the shadow more concentrated could also help.

coppolaemilio:
@YuriSizov added some more games (I think those are all we have for now) and improved the layout a bit. Still not sure how to improve the author. I think it looks good like it is right now :(

coppolaemilio:
~~Made the site be up to date (adding beta 11 article and new sponsor)~~
This will be done in separate PRs

YuriSizov:
While you're at it, migration can be removed from the readme :)

YuriSizov (approved):
I made a couple of fixes, but otherwise this is great as is :)

coppolaemilio:
Okay, moved the authors for the coming soon games to the right side so they are in the same place as the rest of the cards.
Feel free to merge whenever you want 👍

YuriSizov:
Thanks!

PR #504, opened 2023-01-13
Add missing newlines at EOF, strip trailing whitespace

akien-mga:
We should add CI checks to ensure this, like we do in other code repos.

I ran the engine repo's `misc/scripts/file_format.sh`, which also removed trailing whitespace.


akien-mga:
Finally got to test, didn't seem to break stuff from a very quick look.

YuriSizov (approved):

YuriSizov:
Thanks!

PR #505, opened 2023-01-13
Make build scripts executable

akien-mga:
With this I can build with docker on Linux using the command in the README.

`build-and-serve.sh` doesn't work though:
- It complains about missing `webrick` (can be solved with `bundle install webrick`)
- But even then when it deploys to localhost I don't seem able to access it from my actual localhost. Not too familiar with accessing websites that are deployed within a docker so it might be a config issue on my end.

Local python server works.


YuriSizov (approved):
I'll have a look at making and-serve option work. It may be that it's just not supported, as the official Jekyll images are designed to be run as builders, so you don't make a persistent VM, you just build with it and then do whatever you want on the host.

But I think it should still be possible, the server command just needs to keep the process running.

YuriSizov:
> `bundle install webrick`

The command seems to be `bundle add webrick`. Hopefully I'll sort it out soon, but if anyone stumbles upon this issue I thought it's better to clarify :)

PR #503, opened 2023-01-13
Update name and pronouns for HP (legacy version)

akien-mga:
I'll do a PR with the same against the new static site.

PR #506, opened 2023-01-13
Update name and pronouns for HP, including in past blog posts (static version)

akien-mga:
Follow-up to #503 for the new static backend.


coppolaemilio:
Thanks! :)

PR #512, opened 2023-01-15
Add "en" to the value on the default html tag

### LINKED ISSUE #510: Generated HTML has empty `lang` attribute
Running `bundle exec jekyll build` and inspecting the generated HTML pages, all of them include `<html lang="">`. 

_layouts/default.html has `<html lang="{{ site.lang }}">` , but there's no `lang` key in _config.yml and instead of a warning it seems to just fall back to an empty string.

coppolaemilio:
Fixes https://github.com/godotengine/godot-website/issues/510

Since we are not using the localization plugin yet, I believe it is better to leave it like this without using any variable.
Once we enable that plugin again we can revert this.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ site.lang }}">
+<html lang="en">
 
 <head>
 	<meta charset="utf-8">


YuriSizov (approved):

YuriSizov:
Thanks!

PR #523, opened 2023-01-16
Add site.baseurl variable to fix OG image previews

coppolaemilio:
The OG image was not formatted properly. Even if it worked in most sites, some others like Twitter were not pulling the images because we were using a relative path instead of an absolute one.

This PR adds a `_config.yml` variable to use in `/rss.xml` and `/_layouts/default.html`.


YuriSizov (approved):

YuriSizov:
Thanks!

PR #529, opened 2023-01-17
Fix redirect template

### LINKED ISSUE #527: Incorrect forwarding from godotengine.org/news/ leads to godotengine.org/https://godotengine.org/blog
It looks like when moving https://godotengine.org/news/ to https://godotengine.org/blog/ the forwarding has been set up incorrectly, so https://godotengine.org/news/ now send you to https://godotengine.org/https://godotengine.org/blog which is just a 404 error.

coppolaemilio:
The redirect plugin started to add a `/` before the url, so while this PR doesn't fix the underlying issue, it does fix the redirects.

Not sure what the origin of the bug is, but this PR will replace the first instance of `/https` with `https` so that we get rid of that first forward slash if it exists. 

Should close: https://github.com/godotengine/godot-website/issues/527

### DIFF
--- a/_layouts/redirect.html
+++ b/_layouts/redirect.html
@@ -1,0 +1,12 @@
+{% assign target = page.redirect.to | replace_first: "/https:", "https:" %}
+<!DOCTYPE html>
+<html lang="en-US">
+  <meta charset="utf-8">
+  <title>Redirecting&hellip;</title>
+  <link rel="canonical" href="{{ target }}">
+  <script>location="{{ target }}"</script>
+  <meta http-equiv="refresh" content="0; url={{ target }}">
+  <meta name="robots" content="noindex">
+  <h1>Redirecting&hellip;</h1>
+  <a href="{{ target }}">Click here if you are not redirected.</a>
+</html>


_layouts/redirect.html
```diff
@@ -0,0 +1,12 @@
...
+<html lang="en-US">
+  <meta charset="utf-8">
+  <title>Redirecting&hellip;</title>
+  <link rel="canonical" href="{{ target }}">
+  <script>location="{{ target }}"</script>
+  <meta http-equiv="refresh" content="0; url={{ target }}">
+  <meta name="robots" content="noindex">
+  <h1>Redirecting&hellip;</h1>
+  <a href="{{ target }}">Click here if you are not redirected.</a>
+</html>
```
YuriSizov:
```suggestion
</html>

```

coppolaemilio: 💀 sorry

coppolaemilio:
This template is based on the one provided by the plugin itself: https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect.html

coppolaemilio:
The proper solution was found with #530

PR #530, opened 2023-01-17
Don't shadow base config options, add local dev config

### LINKED ISSUE #527: Incorrect forwarding from godotengine.org/news/ leads to godotengine.org/https://godotengine.org/blog

YuriSizov:
Fixes https://github.com/godotengine/godot-website/issues/527. Supersedes https://github.com/godotengine/godot-website/pull/529.

Seems like our hunch was correct, the redirect plugin uses the `baseurl` setting to generate URLs. We tried to use it as a protocol + hostname base, but the plugin uses it as a base path, suggesting to use `url` for the domain. Furthermore, `baseurl` is a part of the Jekyll's default configuration options, though its usage there is not clear to me.

Either way, seems like a safer solution here is to define `url` instead of `baseurl`, and use that throughout the templates. RSS generation, and the rest still work, and the redirects are fixed now.

I also went ahead and created `_config.development.yml`, so we can define `url` for local development, as well as any future setting we might want to adjust locally. Without that all generated URLs refer to the `https://godotengine.org` domain, which makes it harder to test things locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ To build the website locally, follow these steps:
 2. Clone this repository.
 3. Install the necessary dependencies: `bundle install`.
 4. Build the site: `bundle exec jekyll build`.
+   - Append `--config _config.yml,_config.development.yml` to use the development config with your build.
 
 For simplicity, these two commands are also available as a `build.sh` script in this repository.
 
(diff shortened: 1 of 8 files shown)


coppolaemilio (approved):

coppolaemilio:
Thanks! This was the proper fix to the issue 😅

PR #537, opened 2023-01-20
Blog: add note for articles that might be outdated

### LINKED ISSUE #534: Add a notice at the top of the article page for outdated articles
Our blog can sometimes be treated as a knowledge base because several features have been historically documented there (before being ported to the docs, or instead of that). So old articles sometimes appear as relevant search results in web searches, or are shared throughout the community, while the information in them keeps getting out of date.

While we could go through each article manually and update it, it would make sense to add a general notice to every article older than XX months that it may be potentially outdated. We can probably skip the release notes with that.

As the website is only updated with builds, we could make that notice appear with some JS, but I think it's also okay to update it periodically through new deploys when we publish stuff. We do it regularly enough to refresh old pages in time.

merumelu:
Provides a custom `site.outdated_article_threshold` variable through a Jekyll plugin that is used to check if an article is outdated. This has to be done this way because the template engine does not appear to support any math operations, you can only increment or decrement numbers by 1 :)

The note is shown if a post is in the "news", "progress-report", or "events" category and around 18 months or older at build time, based on suggestion by @Calinou. The wording was also changed from the initial design in #534 to not refer to the time threshold directly.
note

You can override the message by adding `outdated_notice_text: TEXT` to a posts front matter (suggested by @coppolaemilio) and disable the notification altogether by adding `no_outdated_notice: true`.
custom_text

This closes #534.


YuriSizov:
I think you can borrow colors from the yellow notices in the documentation (see https://docs.godotengine.org/en/latest/). They are a bit more pleasant on the eyes. I think we should keep the text straight, without italics, and I would also make the date bolder to stand out more.

merumelu:
> I think you can borrow colors from the yellow notices in the documentation (see https://docs.godotengine.org/en/latest/).

I did but turns out only halfway 😄. The background colors are correct but I forgot to change the text colors as well, whoops.

merumelu:
Okay, now the colors are entirely the same as in the documentation.
ScreenShot
(Light theme omitted because the change isn't really noticable there.

YuriSizov:
I wonder if we should also link to the documentation (just the home page of it) as a suggested action. Like, 

> You can find up-to-date articles about the engine in the [official documentation](https://docs.godotengine.org/en/stable/).

akien-mga:
Looks good! I agree with adding a link to the docs.

Maybe "some or all of its contents" can be simplified to "some of its contents"? It makes the sentence easier to read IMO and the "all" case isn't necessarily something that would apply often (and one can understand "some" as anything between 0% and 100% :)).

merumelu:
Okay, text is now
> This article is from (MONTH), some of its contents might be outdated and no longer accurate.
>
> You can find up-to-date information about the engine in the [official documentation](https://docs.godotengine.org/en/stable/).

Calinou (approved):
Thanks!

PR #546, opened 2023-01-29
Add more fields to RSS items

### LINKED ISSUE #532: Add item author to rss feed
Please can you add author names to items in the rss feed? I believe the correct tag is dc:creator, but I'm not an expert.

coppolaemilio:
This PR adds the missing fields on our RSS feed:
- Category
- guid
- author
- image

This closes https://github.com/godotengine/godot-website/issues/532

### DIFF
--- a/pages/rss.xml
+++ b/pages/rss.xml
@@ -16,12 +16,12 @@ permalink: /rss.xml
 	<item>
 		<title>{{ post.title }}</title>
 		<link>{{ site.url }}{{ post.url }}</link>
-		<description>{{ post.content | xml_escape }}</description>
+		<description>{{ post.excerpt }}</description>
 		{% for category in post.categories %}
 		<category>{{ site.data.categories[category][0].name }}</category>
 		{% endfor %}
 		<guid>{{ site.url }}{{ post.url }}</guid>
-		<author>{{ post.author }}</author>
+		<dc:creator>{{ post.author }}</dc:creator>
 		<pubDate>{{ post.date | date: "%a, %d %b %Y %X +0000" }}</pubDate>
 		<image>{{ site.url }}{{ post.image }}</image>
 	</item>
--- a/pages/rss.xml
+++ b/pages/rss.xml
@@ -17,7 +17,13 @@ permalink: /rss.xml
 		<title>{{ post.title }}</title>
 		<link>{{ site.url }}{{ post.url }}</link>
 		<description>{{ post.content | xml_escape }}</description>
+		{% for category in post.categories %}
+		<category>{{ site.data.categories[category][0].name }}</category>
+		{% endfor %}
+		<guid>{{ site.url }}{{ post.url }}</guid>
+		<author>{{ post.author }}</author>
 		<pubDate>{{ post.date | date: "%a, %d %b %Y %X +0000" }}</pubDate>
+		<image>{{ site.url }}{{ post.image }}</image>
 	</item>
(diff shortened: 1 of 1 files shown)


coppolaemilio:
It seems like the author field should have an email to be valid, but I couldn't find any information about it being a problem if it doesn't contain one. It mostly gets skipped because of the obvious spam issues, but I think it should help to have the author field like we have it here.

coppolaemilio:
I replaced the `<author>` tag by `<dc:creator>` which seems to be the popular way to do it. It seems to be a standard from the Dublin Core Metadata Initiative. Checked some popular sites feeds to adapt to what they were doing. 

Replaced the `description` for the `excerpt` since it seems like we shouldn't be shipping the entire content of the blog post, that's something that the reader should be extract from our posts. This should only be a truncated version of it.

It is now ready again for review.

Calinou (approved):

JohnVeness:
Many thanks for these changes @coppolaemilio , especially the author and image. Personally I like it when feeds contain the entire article text (and over half of the approx 70 feeds I follow do so).

I notice the live site rss.xml hasn't been updated yet so I haven't been able to test the changes. Is it possible to force a rebuild?

coppolaemilio:
Hey @JohnVeness, we haven't pushed the latest changes live yet. Deploys to the website are not automatic at the moment. 
I made a decision based on the RSS `description` specifications for the `<item>` (https://validator.w3.org/feed/docs/rss2.html#hrelementsOfLtitemgt). Since we link to the article, an excerpt seems like the way to go. Also, while leaving the entire post there, we got a lot of warnings because elements such as `iframes` were not valid. 

This will make our rss.xml file smaller, which was one of the biggest volume of traffic in previous versions.

JohnVeness:
Thanks, I understand.

zhangjun-qcqi:
HI @coppolaemilio , I find that we need to add `xmlns:dc="http://purl.org/dc/elements/1.1/"` to the rss file if we want to use `<dc:creator>`, otherwise many rss/xml parsers would simply complain that the rss is invalid xml.

For example, check simplepie: 
https://simplepie.org/demo/?feed=https%3A%2F%2Fgodotengine.org%2Frss.xml

The code could be like:
`<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:dc="http://purl.org/dc/elements/1.1/">`

> I replaced the `<author>` tag by `<dc:creator>` which seems to be the popular way to do it. It seems to be a standard from the Dublin Core Metadata Initiative. Checked some popular sites feeds to adapt to what they were doing.
> 
> Replaced the `description` for the `excerpt` since it seems like we shouldn't be shipping the entire content of the blog post, that's something that the reader should be extract from our posts. This should only be a truncated version of it.
> 
> It is now ready again for review.

JohnVeness:
Yes, I noticed Firefox will no longer show the raw XML (it used to before the dc:creator change), although my usual feed reader isn't complaining.
Screenshot_20230201-085836_1

coppolaemilio:
I tested the RSS feed on many readers and it was working properly. Haven't tried opening it in Firefox like you mentioned but I can see the error message.
I'll add the edit to specify the `xmlns:dc="http://purl.org/dc/elements/1.1/"` 👍 

Thanks!

YuriSizov:
Relevant issue: https://github.com/godotengine/godot-website/issues/548

PR #553, opened 2023-02-09
Overhaul the downloads page

YuriSizov:
Emi and I remade the downloads page to modernize it and refocus the navigation ahead of the Godot 4.0 release. So far the page still only serves a single stable version of the engine. I played around (mostly in my mind) with possibilities to have both Godot 3 and Godot 4 side by side, but it gets tricky because a lot of the content on the page depends on the version we're serving (like instructions and requirements, for example).

I think we can do this overhaul, and then see if we can extend it to let us provide either version with ease. While Godot 4 is all the rage, we still have plenty Godot 3 users, and 3.6 is going to be our LTS version for that.

A gallery of changes:

<details>
<summary>Desktop version</summary>

chrome_2023-02-09_20-15-04
chrome_2023-02-09_20-15-16
chrome_2023-02-09_20-16-56
chrome_2023-02-09_20-17-07
</details>

<details>
<summary>Mobile version</summary>

chrome_2023-02-09_20-19-39
chrome_2023-02-09_20-19-50
chrome_2023-02-09_20-19-58
chrome_2023-02-09_20-20-08
</details> 

<details>
<summary>More desktop</summary>

chrome_2023-02-09_20-23-23
chrome_2023-02-09_20-23-42
image
image
chrome_2023-02-09_20-37-21

</details>


Calinou:
The "Download Godot for (platform)" text seems too large to me, but the rest looks great :slightly_smiling_face:

coppolaemilio:
I think the GitHub section can be improved, but it doesn't have to be on this PR since this is already better than the current version. 
I'll test tomorrow 👍

coppolaemilio:
@YuriSizov what do you think about moving the requirements section after the other sections? I feel like it isn't as important as the other ones:
move

YuriSizov:
> what do you think about moving the requirements section after the other sections?

Well, I specifically brought it up because I think it's more relevant than others :) Requirements are pretty short, but important, especially when we start getting into the 4.0 territory and Vulkan-related issues, and whatnot.

I feel like the main idea from your initial rework (that I tried to follow) is that we give the most relevant download at the top. This reduces the importance of the other downloads significantly, as well as that of the platform switcher (which is why I moved it below as well). While requirements are still relevant even to the most important download options shown at the top. And instructions are just a short scroll away.

Speaking of importance, maybe we should ensure that on mobile the instructions block comes before the all downloads block (can probably just achieve it with a reverse order CSS setting).

> I think the GitHub section can be improved, but it doesn't have to be on this PR since this is already better than the current version.

Yeah, I decided not to touch it. It does its thing, though just like you I'm not entirely sure what thing that is :D

coppolaemilio (approved):

coppolaemilio:
Sounds good, thanks!

YuriSizov:
Let's have some fun 👀

Issue #214, opened 2020-12-17
Allow linking to headers on news pages and the Features page


Calinou:
On news pages and static pages like [Features](https://godotengine.org/features), we should add anchor link icons that appear when a heading is hovered. When the anchor icon is clicked, it navigates to an anchor link that can then be shared to others for easier referencing. For example: https://godotengine.org/features#xrsupport

ampsteric:
hey! , I would be happy to help.
Please assign me this issue under hacktoberfest.

nsundriyal62:
please assign this issue to me !

Ymanawat:
@Calinou I tried to do that and created a pull request, please review it.

JohnVeness:
I think this can be closed as completed...

PR #556, opened 2023-02-12
Anchor links to headings

Ymanawat:
- Made Anchor link icons appear when a heading hovers.
- Generates the id of headings on the page and links that id to headings.
- Generates a Link to that section by clicking the anchor icon.
- Don't make an anchor link to a heading that redirects to another web page, like a blog.
Issue : #214 

https://user-images.githubusercontent.com/81252768/218298143-7df8f50f-c707-457e-92ea-0a5e0c469f28.mp4


YuriSizov:
I would prefer we looked into a build time solution to this.

coppolaemilio:
I also think this was something more for text-heavy pages or blog posts, so I wouldn't add it to every page like it is shown here on the video.
Most of the time, this is needed to link to a specific part of an article (like the downloads section of a beta release).

Ymanawat:
That makes sense to me. Though I thought about not doing it like that, but the docs also have these permalinks under every heading, so I kept it that way. 
I don't want to do it separately for each page, that's why I imported it to default, but I can do it on specific pages if needed. Is there anything i can do about this?

Calinou:
> I don't want to do it separately for each page, that's why I imported it to default, but I can do it on specific pages if needed. Is there anything i can do about this?

You could apply the solution only if the current page's URL contains `/article/`.

YuriSizov:
Since articles are generated as a Jekyll collection from MD files, there should be an opportunity to hook into the generator and add the necessary HTML to the page. (And then add the CSS for the icons to only appear on hover).

Ymanawat:
> You could apply the solution only if the current page's URL contains `/article/`.


It could be done simply by importing the .js file in the article.html.
@YuriSizov Are you thinking of creating a plugin for adding anchor links to the heading in the template and letting them generate from the MD file? I think I can do it.

YuriSizov:
Yes, it would be preferable to generate these links during build time. I'm not sure how pluggable is Jekyll, and it has disappointed me before, but I hope that this is doable.

Ymanawat:
@YuriSizov @Calinou I managed to do it. I removed the `{{content}}` from the article layout and included my own HTML liquid template, which only modifies headings and adds anchors. I added CSS to the article layout itself. Is it going to work?

YuriSizov:
I don't know, update your PR and we'll see :)

YuriSizov:
So apparently (as noted in https://github.com/godotengine/godot-website/pull/574) Jekyll generates IDs for headings in articles. So we don't need a complex solution, I guess. Only a way to add an icon on hover to copy a permalink. This can be done in JS at this point.

Sorry for the runaround.

Ymanawat:
>Jekyll generates IDs for headings in articles


Yes before I didn't know that so in the JS file I was generating the ids but then in Template, I'm not generating IDs but I'm changing `<h2 id="whats-new">What’s new</h2>` to `<h2 id="whats-new"> What’s new <a href="#whats-new" class="anchor-link" >🔗</a></h2>` so that the link got generated in this way and that link icon will point to this section. Should I revert back to that JS solution and remove the generation part and link directly?

YuriSizov:
Yeah, let's go with a JS solution, sorry again.

Ymanawat:
It's okay, Do we want to copy that link on click or keep it as in the above video so that the user will right-click and select copy link address from the more options?

YuriSizov:
Copying on click would be more useful

coppolaemilio:
Thank you for the PR @Ymanawat!

PR #563, opened 2023-02-20
Add a CI workflow for PRs

YuriSizov:
This should allow us to do a quick check that the PR branch compiles at least. More checks can be added in future.


YuriSizov:
Seems to work just fine 😊

PR #569, opened 2023-02-21
Overhaul the Features page

YuriSizov:
This is the final page that is in dear need of an overhaul, so here it is. My goal was to refresh the look and the content without using existing media, since it is mostly outdated or not that pretty. This makes this page a bit text heavy, though in my experience this is pretty typical for such pages. The idea is that the pitch version is on the home page, and here we go in a bit more detail.

That said, we do have some materials in works which we can integrate in future, so it's less monotonous. Until then, this is purely for content. I've tried to write the copy, but you will probably want to tweak or replace it. It is written with Godot 3 in mind, but it should be adaptable to Godot 4 once we are ready to launch.

https://user-images.githubusercontent.com/11782833/220431606-00613e8f-1cfb-4123-ac45-7e11a154758e.mp4

This PR also contains some tweaks to the footer, and a small adjustment to colors of the donation page.


YuriSizov:
Based on initial feedback, adjusted some visuals as well as spacing between cards and other elements.

image
chrome_2023-02-21_22-22-55

There are still a couple of things to change, but that's for tomorrow.

YuriSizov:
Taa-daa!

chrome_2023-02-22_03-34-08

YuriSizov:
With Emi's blessing, we're going live :eyes:

PR #573, opened 2023-02-24
Promote Play Store version of the Android editor

### LINKED ISSUE #571: Add Android editor on Google Play to Download page
We now distribute the Android editor officially on Google Play:

https://play.google.com/store/apps/details?id=org.godotengine.editor.v3
https://play.google.com/store/apps/details?id=org.godotengine.editor.v4

For our Download page, we should probably promote this one as the main download option (but still give access to the APK on TuxFamily for those who don't use/want to use Google Play).

We might want to make room for a future F-Droid distribution we'd want to link to as an option.

CC @m4gr3d

YuriSizov:
Closes https://github.com/godotengine/godot-website/issues/571. I made very simple changes, but we can sure expand on the information for this page if we need to.

chrome_2023-02-24_16-08-12
image

### DIFF
--- a/_layouts/download.html
+++ b/_layouts/download.html
@@ -339,7 +339,11 @@ layout: default
 			{% for download in featured_downloads %}
 
 			{% capture download_url %}
-				https://downloads.tuxfamily.org/godotengine/{{site.stable_version}}/{% if download.mono %}mono/{% endif %}Godot_v{{site.stable_version}}-stable_{{ download.slug }}
+				{% if download.custom %}
+					{{download.custom}}
+				{% else %}
+					https://downloads.tuxfamily.org/godotengine/{{site.stable_version}}/{% if download.mono %}mono/{% endif %}Godot_v{{site.stable_version}}-stable_{{ download.slug }}
+				{% endif %}
 			{% endcapture %}
 
 			<div class="main-download">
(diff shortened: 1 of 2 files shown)


m4gr3d:
We should remove the link for the export templates for now since it's unsupported on the Android port at the moment.

YuriSizov:
> We should remove the link for the export templates for now since it's unsupported on the Android port at the moment.

Done! And changed the text for the buttons too based on the feedback from @akien-mga

coppolaemilio:
Since we already say Android on the title and the icon, should we drop it in the subtitle of the buttons? otherwise it feels too long

image

YuriSizov:
@coppolaemilio This would be true for any platform. We can sure tweak the look, but it's not important in this PR.

coppolaemilio (approved):

m4gr3d:
In the `Instructions` section, I'd rephrase the sentence
```
the experience may be frustrating
```

to 
```
the experience may be suboptimal
```

The current version seems like it's trying to scare users away from trying it out :)

YuriSizov:
I think the original intent was to be understanding 🙃  But I can change that!

akien-mga (approved):
Looks good to me. I'd do the slight language change suggested by m4gr3d too, it's better.

YuriSizov:
Done and merged :)

PR #575, opened 2023-02-24
blog: Add "What's new in C# for Godot 4.0"

raulsntos:


collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- Merge .NET 6 branch with master ([GH-64089](https://github.com/godotengine/godot/pull/64089)).
+- C#: Generators ignore _Ready method in partial classes from another generators ([GH-66597](https://github.com/godotengine/godot/issues/66597)).
+- Allow for a way to ensure some source generators run before / after others ([GH-57239](https://github.com/dotnet/roslyn/issues/57239)).
+
+## The new Variant type
+
+Previous versions of Godot have used `System.Object` to represent Godot's Variant type, which is an union of types supported by Godot. The problem is that `System.Object` is the base type for all .NET types, including types that are not supported by Godot. This means that the Godot API allowed using types that would fail on runtime since we wouldn't be able to convert it to a type that Godot would understand.
```
coppolaemilio:
```suggestion
Previous versions of Godot have used `System.Object` to represent Godot's Variant type, which is a union of types supported by Godot. The problem is that `System.Object` is the base type for all .NET types, including types that are not supported by Godot. This means that the Godot API allowed using types that would fail on runtime since we wouldn't be able to convert it to a type that Godot would understand.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+- C#: Add grouping attributes for properties ([GH-64742](https://github.com/godotengine/godot/pull/64742)).
+- C#: Preserve order of exported fields/categories ([GH-64852](https://github.com/godotengine/godot/pull/64852)).
+- Support explicit values in flag properties, add C# flags support ([GH-59327](https://github.com/godotengine/godot/pull/59327)).
+
+## Improved C# documentation
+
+We have updated the [C# differences documentation](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html) to include tables with the C# equivalent of common GDScript API. Sometimes an equivalent in the Godot C# API does not exist because there are existing methods in the .NET Base Class Library that we would like you to use instead, this reduces the maintainability burden since we don't have to provide that API in Godot and helps users discover C# API that they can use in non-Godot projects.
+
+The new equivalence tables go in as much detail about what's available, and will hopefully be a great tool for developers looking to port GDScript code to C#. Let us know if this can be improved!
```
coppolaemilio:
```suggestion
The new equivalence tables go into as much detail about what's available, and will hopefully be a great tool for developers looking to port GDScript code to C#. Let us know if this can be improved!
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+If this API changes in a future version and the new method signature looks like this:
+
+```csharp
+public void MyMethod(int number = 42);
+```
+
+Even though the parameter is optional, the method is still a different signature therefore libraries built against the previous version that invoke `MyMethod` won't be able to find it.
+
+We will keep trying to get better at avoiding breaking compatibility, but keep in mind there's a number of APIs that we are planning to change in 4.1 that will result in breaking compatibility changes. In order to avoid issues make sure your game project as well as the libraries it uses are all referencing the same _GodotSharp_ version.
```
coppolaemilio:
```suggestion
We will keep trying to get better at avoiding breaking compatibility, but keep in mind there are a number of APIs that we are planning to change in 4.1 that will result in breaking compatibility changes. In order to avoid issues make sure your game project as well as the libraries it uses are all referencing the same _GodotSharp_ version.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- [Announcing .NET 6 — The Fastest .NET Yet](https://devblogs.microsoft.com/dotnet/announcing-net-6/).
+
+### Full AOT and trimming support
+
+This can greatly reduce binary size and enable further performance optimizations.
+
+In order to support this, we'll need to ensure that our libraries are prepared. Dynamically accessed types can't be trimmed, so we need to avoid reflection as much as possible. Some platforms don't support JIT, so we definetely need to support AOT. We'll be working on this as we add support for more platforms in future 4.x releases.
```
coppolaemilio:
```suggestion
In order to support this, we'll need to ensure that our libraries are prepared. Dynamically accessed types can't be trimmed, so we need to avoid reflection as much as possible. Some platforms don't support JIT, so we definitely need to support AOT. We'll be working on this as we add support for more platforms in future 4.x releases.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
+---
+title: "What's new in C# for Godot 4.0"
+excerpt: "All about the new changes that Godot 4.0 brings to C# developers."
```
YuriSizov:
```suggestion
excerpt: "All about changes and improvements that Godot 4.0 offers to C# developers."
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+categories: ["progress-report"]
+author: Raul Santos
+image: /storage/blog/covers/whats-new-in-csharp-for-godot-4-0.webp
+date: 2023-02-25 17:00:00
+---
+
+The Godot 4.0 release includes a lot of new features and improvements, check the release announcement for an overview on all the changes. In this article, we'll go into more detail about the C# specific changes.
```
YuriSizov:
```suggestion
The Godot 4.0 release contains [multiple changes](/article/dev-snapshot-godot-4-0-beta-1) to almost every area of the engine, and first-party support for C# and .NET is no exception. The main focus of this release has been on modernizing the runtime environment and updating the C# API to be more in line with the engine core. We took care not to sacrifice that aspect of familiarity that many of you seek when picking C# as your main scripting language.

In this article, we'll go into more detail about each of the most important changes.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+  - [Mobile and web support](#mobile-and-web-support)
+  - [Full AOT and trimming support](#full-aot-and-trimming-support)
+  - [GDExtension support](#gdextension-support)
+  - [Improve Source Generators extensibility](#improve-source-generators-extensibility)
+  - [Editor unification](#editor-unification)
+  - [Godot as a library](#godot-as-a-library)
+
+<div class="card card-warning">
+  <p><strong>⚠️ Warning</strong></p>
+  <p>Projects created with a previous unstable version of Godot 4.0 will need you to delete the <code>res://.godot/mono</code> directory before opening them in the stable version of Godot 4.0, to ensure that no older assemblies are used.</p>
```
YuriSizov:
```suggestion
  <p>If you have created a Godot project in one of the earlier preview releases of Godot 4.0, you will need to delete the <code>res://.godot/mono</code> directory before opening it in the stable version of the engine. This is required to ensure that no older assemblies are used.</p>
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+  <p>Godot 3 projects can be converted to Godot 4.0 with the built-in project converter, but due to limitations of how the project converter works using regex, the converted project will likely require lots of manual fixing before it can be opened with Godot 4.0. We don't officially support opening Godot 3 C# projects with Godot 4.0 and recommend to create a new project from scratch.</p>
```
YuriSizov:
```suggestion
  <p>Godot 3 projects can be converted to Godot 4.0 with the built-in project converter. Due to limitations of how the project converter works, the converted project will likely require lots of manual fixing before it can be opened with Godot 4.0. Because of numerous changes between the two versions you may want to create a new project from scratch instead of porting it.</p>
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+</div>
+
+## Runtime changes
+
+The main change for C# support is that we moved away from the Mono SDK and we now use the **.NET SDK** to embed the .NET runtime, this means we use the CoreCLR runtime for desktop platforms, and in the future we would use the Mono runtime for mobile. Mobile and web support will be added in a future Godot release.
```
YuriSizov:
```suggestion
The main change for C# support is that we moved away from the Mono SDK and we now use the **.NET SDK** to embed the .NET runtime. This means we use the CoreCLR runtime for desktop platforms. Mobile and web platforms are not supported at the moment, and will be worked on for a future Godot 4 release. We are considering using the Mono runtime for the mobile platform.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+With this change, the Mono module is now called **.NET module**, but the only .NET language officially supported is still **C#**. The Godot API is also now targeting .NET 6, the latest .NET LTS release.
+
+For end users, this means they just need to install the latest [.NET SDK](https://get.dot.net) to start working with C# in Godot. The default target for game projects will be .NET 6 allowing you to use the new modern C# features included in .NET that are not easily accessible in Godot 3. It should also be possible to target newer frameworks, although we can't guarantee it will work properly.
```
YuriSizov:
```suggestion
For end users this means that they only need to install the latest [.NET SDK](https://get.dot.net) to start working with C# in Godot 4. The default target for game projects is .NET 6, allowing you to use modern C# features included in .NET; something that are not easily accessible in Godot 3. It should also be possible to target newer frameworks, although we can't fully guarantee it would work properly.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+The C# language version is not tied to the targeted framework which means even in Godot 3, which targets .NET Framework 4.7.2 by default, you could use the latest C#. However, some features require runtime support (such as [Default Interface Methods](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods)). Now that we support targeting .NET 6, all features available in C# 10 are supported.
```
YuriSizov:
```suggestion
In Godot 3, which targets .NET Framework 4.7.2 by default, you could use the latest version of C# just as well, because the language version is not inherently tied to the targeted framework. However, this is not true for all features, as some of them require runtime support (such as [Default Interface Methods](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods)). Now that we officially support targeting .NET 6, all features available in C# 10 are supported.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Targeting the latest frameworks also allows using the newer APIs included in the Base Class Library (e.g.: `DateOnly`, `TimeOnly`, `PriorityQueue`) and benefiting from [performance improvements](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/).
```
YuriSizov:
```suggestion
Targeting the latest frameworks also allows using the newer APIs included in the Base Class Library (e.g. `DateOnly`, `TimeOnly`, `PriorityQueue`) and benefiting from recent [performance improvements](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/).
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+#### Relevant links
+
+- Merge .NET 6 branch with master ([GH-64089](https://github.com/godotengine/godot/pull/64089)).
+- [What's new in .NET 5](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-5).
+- [What's new in .NET 6](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-6).
+- [Announcing .NET 6 — The Fastest .NET Yet](https://devblogs.microsoft.com/dotnet/announcing-net-6/).
+
+## Engine interop with source generators
+
+In the past, Godot has used reflection to communicate with the engine, this has worked great so far but has some limitations. We now use [source generators](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/), which has these benefits:
```
YuriSizov:
```suggestion
In the past, Godot has used reflection to communicate with the engine. This has worked great so far but it comes with some limitations. We now use [source generators](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/), which have the following benefits:
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+- Generated code is generally faster than reflection.
+- Supports **trimming** which reduces the size of the exported assembly and paves the way for full **AOT** support in the future.
+- Allows analyzing user code to **emit diagnostics on compile-time**, which means we can warn you about code that would fail on runtime before you actually execute the game.
```
YuriSizov:
```suggestion
- They support **trimming** which reduces the size of the exported assembly and paves the way for full **AOT** support in the future.
- They allow analyzing user code to **emit diagnostics at compile time**. This means that we can warn you about code that would fail at runtime before you actually execute the game.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+This also means user-types that derive from Godot types (such as `Node` or `Resource`) need to be declared `partial`.
```
YuriSizov:
```suggestion
This change also means that user types which derive from native Godot types (such as `Node` or `Resource`) need to be declared `partial`.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- Merge .NET 6 branch with master ([GH-64089](https://github.com/godotengine/godot/pull/64089)).
+- C#: Generators ignore _Ready method in partial classes from another generators ([GH-66597](https://github.com/godotengine/godot/issues/66597)).
+- Allow for a way to ensure some source generators run before / after others ([GH-57239](https://github.com/dotnet/roslyn/issues/57239)).
+
+## The new Variant type
+
+Previous versions of Godot have used `System.Object` to represent Godot's Variant type, which is a union of types supported by Godot. The problem is that `System.Object` is the base type for all .NET types, including types that are not supported by Godot. This means that the Godot API allowed using types that would fail on runtime since we wouldn't be able to convert it to a type that Godot would understand.
```
YuriSizov:
```suggestion
Previous versions of C# in Godot have used `System.Object` to represent Godot's Variant type, which is a special container for various data types supported by the engine. The problem with that is that `System.Object` is the base type for all .NET types, including types that are not supported by Godot. This means that the Godot API allowed using types that would fail at runtime since we wouldn't be able to convert it to a type that Godot would understand.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Now we have a dedicated `Variant` struct that represents all the types that Godot supports, this means our API is more strict on which types you can use so you get errors on compile-time before executing your game.
```
YuriSizov:
```suggestion
Now we have a dedicated `Variant` struct that represents all the types that Godot supports. This means our API is more strict about which types you can use, and you will get errors at compile time before executing your game.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+The way the `Variant` struct is implemented also avoids boxing value types (such as `int` and `float`) in cases where it was unavoidable in Godot 3.
+
+When using `System.Object`, users have been able to take advantage of .NET features such as pattern matching and generics, the new `Variant` struct no longer allows to use those features directly. Pattern matching can be replaced with checking the `Variant.VariantType` property and generics can be supported using the `[MustBeVariant]` attribute.
```
YuriSizov:
```suggestion
When using `System.Object`, users have been able to take advantage of .NET features such as pattern matching and generics. The new `Variant` struct no longer allows to use those features directly. Pattern matching can be replaced with checking the `Variant.VariantType` property, while generics can be supported using the `[MustBeVariant]` attribute.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- C#: Represent Variant as its own type instead of System.Object ([GH-3837](https://github.com/godotengine/godot-proposals/issues/3837)).
+- Merge .NET 6 branch with master ([GH-64089](https://github.com/godotengine/godot/pull/64089)).
+- C#: Optimize Variant conversion callbacks ([GH-68310](https://github.com/godotengine/godot/pull/68310)).
+
+## Collections changes
+
+Just like in 3.x, the `Array` and `Dictionary` collections are supported in C# by the types defined in the `Godot.Collections` namespace. Packed arrays are a special type of array, the equivalent of 3.x Pool arrays, in C# the equivalent is C# arrays (e.g.: `byte[]`).
```
YuriSizov:
```suggestion
Just like in 3.x, `Array` and `Dictionary` collections are supported in C# through the types defined in the `Godot.Collections` namespace. Godot's packed arrays, a special type of arrays optimized for memory usage, don't have their own representation in C#, and normal C# arrays should be used (e.g. `byte[]`).
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Godot APIs are only able to understand these Godot collections, which means anything that interops with Godot's native side needs to use one of the supported collection types.
+
+In 3.x, the non-generic collections used `System.Object` like .NET non-generic collections. However, since we now have a [`Variant`](#the-new-variant-type) type, Godot C# collections now implement the generic collection interfaces where `T` is `Variant`. This means that the non-generic collections now only support Variant-compatible types, so you'll get compile errors when using an invalid type rather than a runtime exception.
```
YuriSizov:
```suggestion
In 3.x, the non-generic collections used `System.Object`, just like .NET non-generic collections. However, thanks to the new [`Variant`](#the-new-variant-type) type, Godot C# collections now implement the generic collection interfaces where `T` is `Variant`. This means that the non-generic collections now only support Variant-compatible types, so you'll get compile errors when using an invalid type rather than a runtime exception.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+The generic collections will also validate that the generic `T` parameter is one of the Variant-compatible types since it's annotated with the new `[MustBeVariant]` attribute.
```
YuriSizov:
```suggestion
The generic collections will also validate that the generic `T` parameter is one of the Variant-compatible types, since it's annotated with the new `[MustBeVariant]` attribute.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Godot collections now also implement more utility methods that expose similar functionality to what can be done in other languages. Using these instance methods is faster than Linq because it avoids marshaling every item in the collection.
```
YuriSizov:
```suggestion
Godot collections now implement more utility methods that expose similar functionality to what can be done in other supported languages. Using these instance methods is generally faster than Linq because it avoids marshaling every item in the collection. On top of that, collections also provide an extra utility method, `MakeReadOnly`. It allows you to freeze the collections so modifying them is no longer possible.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Another new API is the `MakeReadOnly` method, which allows you to freeze the collections so modifying them is no longer possible.
```
YuriSizov:
```suggestion
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Besides the Godot collections, the Base Class Library in .NET includes many other collections that can be used. These collections are not supported by Godot but they can still be useful in some scenarios (i.e. pure logic that doesn't interface with the engine).
+
+In 3.x, some .NET collections were supported in very specific scenarios (such as exporting properties). This is no longer supported in 4.0 since it required reflection. Support for .NET collections may be re-introduced in the future but it will have to rely on source generators instead of reflection.
```
YuriSizov:
```suggestion
You are not limited to using Godot collections in your own internal logic, of course. The Base Class Library in .NET includes many other collections that can be used. These collections are not going to work with Godot's API but can still be useful. In 3.x, some .NET collections were supported in very specific scenarios (such as exporting properties). This is no longer supported in 4.0 since it required reflection. Support for .NET collections may be re-introduced in the future, implemented with source generators instead of reflection.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+#### Relevant links
+
+- Merge .NET 6 branch with master ([GH-64089](https://github.com/godotengine/godot/pull/64089)).
+- C# Export on List not possible anymore ([GH-70298](https://github.com/godotengine/godot/issues/70298)).
+- Sync C# Array with Core ([GH-71786](https://github.com/godotengine/godot/pull/71786)).
+- Sync C# Dictionary with Core ([GH-71984](https://github.com/godotengine/godot/pull/71984)).
+
+## Signals as events
+
+Godot signals are a great way to implement the Observer pattern. In C#, it's possible to use Godot signals by calling the `Connect` and `EmitSignal` methods, but C# developers are more used to C# events as they feel more idiomatic.
```
YuriSizov:
```suggestion
Godot signals are a great way to implement the Observer pattern. It's possible to use Godot signals by using the common engine API, such as the `Connect` and `EmitSignal` methods. But C# developers are more used to using events, as they are more idiomatic.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+For Godot 4.0, we now generate C# events for all Godot signals, which allows developers to use the event syntax that they are used to and love. As a consequence, subscribing to Godot signals using the event syntax is type-safe.
+
+In contrast with normal C# events, these generated events differ from normal C# events in two ways:
```
YuriSizov:
```suggestion
These generated events differ from normal C# events in two ways:
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+* These events are disconnected from automatically upon freeing the Godot type that contains the signal.
+* Using the `Invoke` method to raise the event is not allowed.
+
+We are looking into ways to allow raising events in a type-safe way, so keep an eye out for that in the future.
+
+For more information about how to use Signals in C#, take a look at the new documentation page about [_C# Signals_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_signals.html).
```
YuriSizov:
```suggestion
For more information on how to use signals in C#, take a look at the new documentation page about [_C# Signals_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_signals.html).
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+- Fix C# bindings after recent breaking changes ([GH-37050](https://github.com/godotengine/godot/pull/37050)).
+- Create a dedicated "C# Signals" page ([GH-6643](https://github.com/godotengine/godot-docs/pull/6643)).
+
+## Callable support
+
+In Godot 4, we introduce the new `Callable` type that represents a method in an object instance or a standalone function, this may sound familiar because C# already supports a similar concept with delegate types, the `Action` and `Func` types, and lambdas.
+
+In order to support interoperability with the engine, Godot's C# API implements a `Callable` type that can be created from an `Action` or `Func`. This allows users to use C# lambdas with Godot API.
+
+`Callable`s can be invoked by using the `Call` or `CallDeferred` methods. We currently don't support binding values to the parameters in C# but this shouldn't be a problem when using C# lambdas since they can use closures to support this scenario.
```
YuriSizov:
```suggestion
In Godot 4 we introduce the new `Callable` type that represents a method in an object instance or a standalone function. This may sound familiar, because C# already supports a similar concept with delegate types, the `Action` and `Func` types, and lambdas.

In order to support interoperability with the engine, Godot's C# API implements a `Callable` type that can be created from an `Action` or a `Func`. This also allows developers to use C# lambdas with Godot API.

`Callable`s can be invoked by using the `Call` or `CallDeferred` methods. We currently don't support binding values to the parameters in C#. This shouldn't be a problem when using C# lambdas since they can use closures to support this scenario.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Int and Float changes
+
+Godot uses the name `INT` and `FLOAT` to mean 64-bit integer and floating types even in Godot 3, but in C# we were marshaling those types as 32-bit C# types `int` and `float`, which led to [marshaling issues](https://github.com/godotengine/godot/issues/39609) and potential precision loss.
+
+We have fixed this in Godot 4.0 and now we use the same bit-ness as the engine in Godot APIs. This means some APIs have changed from `int` to `long` and from `float` to `double`.
+
+For vector types, Godot uses 32-bit floats by default (this can be changed by building the engine with `--precision=double`). This may lead to some inconvenient situations where values must be converted to `float` before using them with vectors, the most common one being the `delta` parameter in `_Process` and `_PhysicsProcess`.
+
+The `Mathf` API has also been updated to support both `float` and `double` overloads which hopefully reduces friction with these changes.
```
YuriSizov:
```suggestion
In Godot API the names `INT` and `FLOAT` are used to mean 64-bit integer and floating types, respectively. This is true even in Godot 3, but in C# we used to marshal those types as 32-bit C# types `int` and `float`. This led to [marshaling issues](https://github.com/godotengine/godot/issues/39609) and potential precision loss. This has been fixed in Godot 4.0, as now we use the same bit-ness as the engine does. This means some APIs have changed from `int` to `long` and from `float` to `double`.

For vector types, Godot uses 32-bit floats by default (this can be changed by building the engine with `--precision=double`). This may lead to some inconvenient situations where values must be converted to `float` before using them with vectors. This limitation also applies to some engine core APIs, the most common one being the `delta` parameter in `_Process` and `_PhysicsProcess`. The `Mathf` API has been updated to support both `float` and `double` overloads, which should reduce friction with these changes.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+- C#: Assume 64-bit types when type has no meta ([GH-65168](https://github.com/godotengine/godot/pull/65168)).
+- Reduce the amount of casting required for floating points in C# ([GH-5403](https://github.com/godotengine/godot-proposals/issues/5403)).
+- C#: Add float and double overloads to Mathf ([GH-71583](https://github.com/godotengine/godot/pull/71583)).
+
+## Renames
+
+A lot of types and members have been renamed in Core with the intention to make names clearer. In C# we took the opportunity to also rename a bunch of APIs in order to more closely follow the [.NET naming guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines).
+
+The [_Capitalization Conventions_](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) guidelines indicate that C# should use PascalCase for all identifiers except parameter names (including acronyms over two letters in length). As a result, some types have been renamed: `CPUParticles2D` is now `CpuParticles2D`, `DTLSServer` is now `DtlsServer`.
```
YuriSizov:
```suggestion
The [_Capitalization Conventions_](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) guidelines indicate that C# should use PascalCase for all identifiers except parameter names (including acronyms over two letters in length). As a result, some types have been renamed (e.g. `CPUParticles2D` is now `CpuParticles2D`, `DTLSServer` is now `DtlsServer`).
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Exported property improvements
+
+A lot of highly requested features have been implemented in Godot to improve exported properties and the inspector. These features are also supported in C#.
+
+In 4.0, you are now able to export properties of type `Node` or any derived type directly, without the need to export a `NodePath` and retrieve it manually.
+
+It's also now possible to group properties with the new `[ExportCategory]`, `[ExportGroup]` and `[ExportSubgroup]` attributes. Take a look at the updated [_C# Exports_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_exports.html) documentation page to learn more about the new attributes.
+
+Support for flag enums (annotated with the `[System.Flags]` attribute) has also been improved and now show up as checkboxes in the inspector.
```
YuriSizov:
```suggestion
Support for flag enums (annotated with the `[System.Flags]` attribute) has also been improved. Exported properties for such enums now show up as checkboxes in the Inspector.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+Due to [engine interop improvements](#engine-interop-with-source-generators) we now use source generators instead of reflection. This means that the attributes are not retrieved on runtime, instead the source generators generate different code based on the attributes found with source analysis. This makes it much more difficult to support custom user attributes derived from the Godot attributes. With our current source generator implementation, derived attributes have no effect on user scripts. So, to prevent confusion, we've sealed the attributes to disallow inheriting from them.
+
+We are looking into ways to support extending our source generators to allow custom user attributes to affect scripts, so keep an eye out for that in the future.
```
YuriSizov:
```suggestion
Due to [engine interop improvements](#engine-interop-with-source-generators) we now use source generators instead of reflection. This means that the attributes are not retrieved at runtime. Instead the source generators generate different code based on the attributes found from the source analysis. This makes it much more difficult to support custom user attributes derived from the Godot attributes. With our current source generator implementation, derived attributes have no effect on user scripts.

To prevent a possible confusion, we've sealed the attributes to disallow inheriting from them. We are looking into ways to support extending our source generators to allow custom user attributes to affect scripts, so keep an eye out for that in the future.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+In Godot 4.0, GDScript added a bunch of new export annotations. However, we didn't have time to add equivalent attributes in C#. All those annotations can be replicated with the `[Export]` attribute using the right combination of `PropertyHint` and `HintString` values, until we implement these convenient attributes in a future version of Godot 4.x.
```
YuriSizov:
```suggestion
In Godot 4.0,  a bunch of new export annotations were added to GDScript. However, we didn't have time to add equivalent attributes in C#. All those annotations can be replicated with the `[Export]` attribute using the right combination of `PropertyHint` and `HintString` values, until we implement these convenient attributes in a future version of Godot 4.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+As for [custom resources](#global-classes), we were unable to finish this work on time for 4.0, but we're actively working on it.
```
YuriSizov:
```suggestion
The much anticipated support for exporting [custom resources](#global-classes) hasn't made it in time, unfortunately, and is still in development, but you can expect it to arrive very soon.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+- C#: Add grouping attributes for properties ([GH-64742](https://github.com/godotengine/godot/pull/64742)).
+- C#: Preserve order of exported fields/categories ([GH-64852](https://github.com/godotengine/godot/pull/64852)).
+- Support explicit values in flag properties, add C# flags support ([GH-59327](https://github.com/godotengine/godot/pull/59327)).
+
+## Improved C# documentation
+
+We have updated the [C# differences documentation](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html) to include tables with the C# equivalent of common GDScript API. Sometimes an equivalent in the Godot C# API does not exist because there are existing methods in the .NET Base Class Library that we would like you to use instead, this reduces the maintainability burden since we don't have to provide that API in Godot and helps users discover C# API that they can use in non-Godot projects.
+
+The new equivalence tables go into as much detail about what's available, and will hopefully be a great tool for developers looking to port GDScript code to C#. Let us know if this can be improved!
```
YuriSizov:
```suggestion
A lot of Godot tutorials and tips you can find online contain samples of GDScript code. To help you with interpreting them into C# equivalents the official documentation has a [C# differences article](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_differences.html), which has also been updated. It now features a complete set of tables with the C# equivalent for every common GDScript API. Sometimes a Godot C# equivalent does not exist because there are more suitable methods in the .NET Base Class Library.

The new equivalence tables try to encompass every possible case, and should be a great tool for developers looking to port GDScript code to C#. Let us know if this can be improved!
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+- C#: Add table with equivalent Dictionary methods ([GH-6676](https://github.com/godotengine/godot-docs/pull/6676)).
+- Add Color section to C# differences page ([GH-6679](https://github.com/godotengine/godot-docs/pull/6679)).
+
+## NuGet packages
+
+With the Godot 4 release we have also started publishing the Godot assemblies to [NuGet.org](https://www.nuget.org/packages/GodotSharp). This allows developers to create third-party libraries that reference Godot assemblies much more easily.
+
+The NuGet packages versioning follows Godot which uses [semver](https://semver.org/). This means, in theory, that there should not be breaking changes between minor or patch versions. In practice, while we try to avoid breaking compatibility as much as possible, there may still be some minor breaking changes from time to time.
+
+How does breaking compatibility affect your game? If your game references a package _MyLibrary_ which was built against _GodotSharp 4.0.0.0_, but your game is using _GodotSharp 4.1.0.0_ which introduces a breaking change, it will cause runtime exceptions, the most common one being `System.MethodNotFoundException`.
```
YuriSizov:
```suggestion
With the Godot 4 release we also start publishing the Godot assemblies to [NuGet.org](https://www.nuget.org/packages/GodotSharp). This allows developers to create third-party libraries that reference Godot assemblies with more ease and convenience.

The NuGet packages versioning follows Godot which is similar to [semver](https://semver.org/). This means, in theory, that there should not be breaking changes between minor or patch versions. In practice, while we try to avoid breaking compatibility as much as possible, there may still be some minor breaking changes from time to time.

How does breaking compatibility affect your game? If your game references a package _MyLibrary_ which has been built against _GodotSharp 4.0.0.0_, but your game is using _GodotSharp 4.1.0.0_ which introduces a breaking change, it will cause runtime exceptions, the most common one being `System.MethodNotFoundException`.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+public void MyMethod();
+```
+
+If this API changes in a future version and the new method signature looks like this:
+
+```csharp
+public void MyMethod(int number = 42);
+```
+
+Even though the parameter is optional, the method is still a different signature therefore libraries built against the previous version that invoke `MyMethod` won't be able to find it.
```
YuriSizov:
```suggestion
Now, let's assume this API changes in a future version and the new method signature looks like this:

```csharp
public void MyMethod(int number = 42);
```

Even though the parameter is optional, the method now has a different signature. Therefore libraries built against the previous version that invoke `MyMethod` won't be able to find it.
```

File: collections/_article/whats-new-in-csharp-for-godot-4-0.md [hunk 38]
```diff
@@ -0,0 +1,357 @@
...
+
+We will keep trying to get better at avoiding breaking compatibility, but keep in mind there are a number of APIs that we are planning to change in 4.1 that will result in breaking compatibility changes. In order to avoid issues make sure your game project as well as the libraries it uses are all referencing the same _GodotSharp_ version.
```
YuriSizov:
```suggestion
Godot implements various systems to preserve compatibility and allow for smooth transition, even when we have to change some fundamental things. You should expect future releases of Godot 4 to work with your Godot 4.0 project, but we still leave the room for the inevitable. In case of completely incompatible changes introduced in minor releases, we promise to provide clear migration instructions and possibly tools for automated conversion.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+## IDE support
+
+With the move to the .NET Core CLR runtime, Godot gained support to use the official C# debugger. This means IDEs that support C# and official C# extensions support debugging Godot projects out of the box. Although configuring your IDE may require a few more steps at the moment.
+
+Other features of Godot's IDE extensions for C#, such as our Roslyn completion providers, are planned to be moved out of the extensions and into NuGet packages so that all IDEs can benefit without requiring a dedicated extension.
+
+The [_Configuring an external editor_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor) section of the C# basics documentation page has yet to be updated to explain how to setup with Godot 4.0, this is still very much a work in progress.
```
YuriSizov:
```suggestion
The [_Configuring an external editor_](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_basics.html#configuring-an-external-editor) section of the C# basics documentation page has yet to be updated to explain how to set up your editor for Godot 4.0. Stay tuned for an updated version.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- Make GodotCompletionProviders work via NuGet ([GH-18](https://github.com/godotengine/godot-csharp-visualstudio/issues/18)).
+
+## Future
+
+Some of the features that didn't make it in time for Godot 4.0 and that we already plan to work on:
```
YuriSizov:
```suggestion
Some features didn't make it in time for Godot 4.0, but are already on our roadmap for a future release (and some are already in the works).
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+### Global classes
+
+Global classes, also known as named scripts, are classes registered in the editor so they can be used more conveniently. These classes appear in the _Add Node_ and _Create Resource_ dialogs. GDScript supports this feature with the `class_name` syntax.
+
+We tried, but ultimately failed, to bring this feature to C# in time for 4.0, but there's already a PR open so it's very likely it will be ready for the next release, Godot 4.1.
```
YuriSizov:
```suggestion
We tried, but ultimately failed, to bring this feature to C# in time for 4.0. Luckily, there's already a PR open, so it's very likely it is going to be ready for the next minor release, Godot 4.1.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+- Add first-class custom resource support ([GH-18](https://github.com/godotengine/godot-proposals/issues/18)).
+- Allow exporting custom resources from/to GDScript, VisualScript, C#, and PluginScript ([GH-48201](https://github.com/godotengine/godot/pull/48201)).
+- Script-class-aware Inspector & related controls ([GH-62413](https://github.com/godotengine/godot/pull/62413)).
+- Enable QuickOpen to see scripted resources ([GH-62417](https://github.com/godotengine/godot/pull/62417)).
+- Add C# resource export ([GH-72619](https://github.com/godotengine/godot/pull/72619)).
+
+### Mobile and web support
+
+We aim to support C# in all the platforms that Godot is available on. Unfortunately, we were unable to implement support for mobile and web platforms for 4.0, these platforms gained support upstream somewhat recently so we didn't have much time to work on it. The new workloads in .NET 6 should allow us to support the mobile and web platforms really soon, so keep an eye out for that.
```
YuriSizov:
```suggestion
We aim to support C# on all the platforms that Godot is available on. Unfortunately, we were unable to implement support for mobile and web platforms for 4.0. These platforms gained support upstream somewhat recently so we didn't have much time to work on it. The new workloads in .NET 6 should allow us to support the mobile and web platforms really soon, so keep an eye out for that.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- [Announcing .NET 6 — The Fastest .NET Yet](https://devblogs.microsoft.com/dotnet/announcing-net-6/).
+
+### Full AOT and trimming support
+
+This can greatly reduce binary size and enable further performance optimizations.
+
+In order to support this, we'll need to ensure that our libraries are prepared. Dynamically accessed types can't be trimmed, so we need to avoid reflection as much as possible. Some platforms don't support JIT, so we definitely need to support AOT. We'll be working on this as we add support for more platforms in future 4.x releases.
```
YuriSizov:
```suggestion
Support for these features can greatly reduce binary size and enable further performance optimizations. In order to support them, we need to ensure that our libraries are prepared. Dynamically accessed types can't be trimmed, so we need to avoid reflection as much as possible.

Some target platforms don't support JIT, so we are definitely looking into supporting AOT. We'll be working on this as we add support for more platforms in future 4.x releases.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+#### Relevant links
+
+- [App Trimming in .NET 5](https://devblogs.microsoft.com/dotnet/app-trimming-in-net-5/).
+- [Native AOT deployment](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/).
+
+### GDExtension support
+
+The ability to create GDExtensions using C# would enable powerful workflows.
+
+This would allow us to avoid the limitations of a scripting language implementation, such as [relying on file paths](https://github.com/godotengine/godot/issues/15661) to reference C# classes. Users will be able to implement C# types that can be registered in ClassDB and behave as built-in Nodes and Resources.
```
YuriSizov:
```suggestion
The ability to create GDExtensions using C# will enable new and powerful workflows. This would allow us to avoid the limitations of a scripting language implementation, such as [relying on file paths](https://github.com/godotengine/godot/issues/15661) to reference C# classes. Users are going to be able to implement C# types that can be registered in Godot's internal type database, ClassDB, and that would behave as built-in Nodes and Resources.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+GDExtension is still very new and has some limitations that would result in some UX regressions if C# moved away from the scripting language implementation, but it also has the potential to reduce some pain points that users have had in the past. Therefore, in Godot 4.0, C# is still implemented as a scripting language and, in the future, we'll add support to create GDExtensions using C# while keeping the scripting language implementation.
+
+Consuming APIs provided by a GDExtension is also currently unsupported, we'll keep working on bridging the gap between GDExtension and C# in future 4.x releases.
```
YuriSizov:
```suggestion
GDExtension is still very new to the engine, and is subject to change. It also has some limitations that would result in UX regressions, if C# moved away from the scripting language implementation. But it also has the potential to reduce some pain points that users have had in the past. In Godot 4.0, C# is still implemented as a scripting language, but we are planning to add support for GDExtension creation using C#. We will be keeping the scripting language implementation for your regular game scripting needs.

Consuming APIs provided by a GDExtension is also currently unsupported. We'll keep working on bridging the gap between GDExtension and C# in future 4.x releases.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- [Introducing GDNative's successor, GDExtension](https://godotengine.org/article/introducing-gd-extensions/).
+
+### Improve Source Generators extensibility
+
+Allowing users to create their own third-party source generators that can extend what's currently available. For example, allowing our source generators to be affected by custom user-defined attributes.
+
+This is tricky to get right, but we'll be exploring some ideas in future 4.x releases to see how we can provide some extensibility without requiring users to disable and re-implement all of our source generators.
```
YuriSizov:
```suggestion
### Improve source generators extensibility

As mentioned throughout this article, several areas can benefit from allowing users to create their own third-party source generators that can extend what's currently available. For example, allowing our source generators to be affected by custom user-defined attributes.

This is tricky to get right, but we'll be exploring some ideas in future 4.x releases to see how we can provide some extensibility without requiring users to disable and re-implement all of our source generators.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+### Editor unification
+
+We didn't have time to finalize this for 4.0, but in the future the goal is to have a single build that supports C# as well as all the other languages supported in the _standard_ version. This would have no cost to non-C# users since the components required to support .NET projects would be downloaded on-demand when a user needs them.
```
YuriSizov:
```suggestion
Just like Godot 3 before it, Godot 4.0 is going to come with two versions of the editor, one "standard" and one with extra support for C# and .NET. However, the plan for 4.x is to provide one unified editor, that support everything the .NET build currently supports. If you're a GDScript user, this doesn't come at any cost for you. But for maintainers and C# developers it greatly simplifies things and allows for more efficient editor development and testing. To keep things slim and flexible, components required to support .NET projects would be downloaded on-demand when the user needs them.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- From embedding Mono to Godot as a library and the future ([GH-2333](https://github.com/godotengine/godot-proposals/issues/2333)).
+
+### Godot as a library
+
+The concept of Godot as a library and C# being the entry point is something that a lot of users seem to be interested in. We think this could bring many benefits to C# users, it would make it easier to support all the platforms where .NET is available, and stay up-to-date with newer .NET versions.
+
+We started exploring some of this work in 4.0, but ultimately this is very much a work in progress. It's unlikely to be finished any time soon but we'll keep working on it in the future.
```
YuriSizov:
```suggestion
We started exploring this area in 4.0, but it still requires more investigation and work. It's unlikely to be finished any time soon but we'll keep working on it in the future.
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Relevant links
+
+- From embedding Mono to Godot as a library and the future ([GH-2333](https://github.com/godotengine/godot-proposals/issues/2333)).
+
+---
+
+And of course, we are always looking to improve performance and usability of the API as well as fixing all the bugs we can find. We are looking forward to see what you think of the .NET module in Godot 4.0, feedback is welcomed! 🙂
```
YuriSizov:
```suggestion
As you can see, a lot of work has been done, and more work is in the pipeline. We are constantly looking into more way to improve performance and usability of the API, and we keep fixing all the bugs that you can find. We are looking forward to see what you think of the .NET module in Godot 4.0. Don't hesitate to reach out with your feedback! 🙂
```

collections/_article/whats-new-in-csharp-for-godot-4-0.md
```diff
@@ -0,0 +1,346 @@
...
+For example, take a look at this method signature:
+
+```csharp
+public void MyMethod();
+```
+
+Now, let's assume this API changes in a future version and the new method signature looks like this:
+
+```csharp
+public void MyMethod(int number = 42);
```
YuriSizov, line 255:
```suggestion
public void MyMethod(int number = 42);
```

```

Oops

YuriSizov: Can't fix this on GitHub, will fix locally

YuriSizov (approved):
Done with proofreading. Amazing summary! Needs a new cover image, but otherwise it's good to go.

Edit: Tested it locally as well.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: This paragraph wasn't so much about updating Godot game projects, it's more about using a NuGet package that targets an older version of the _GodotSharp_ assembly.

In this case, if Godot breaks compat, the library will throw exceptions so providing automated conversion or migration instructions are pretty much impossible. The only options that the user has are:

- Stop using the library entirely.
- Wait for the library to release an update that argets a newert version of _GodotSharp_.

So what I wanted to convey with my paragraph was that we'll try to avoid breaking compatibility, so that libraries released _today_ are still usable in a future version of Godot. We obviously can't promise that breaking changes will never happen, but I do hope that we can avoid breaking changes as much as possible (especially after 4.1).

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: There's two spaces before `a bunch of new [...]`.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: I'm fine with the change, but I wanted to ask if there's a reason we're avoiding describing `Variant` as an union.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: "something that are" doesn't sound right to me since "something" is singular. The subject of "are not easily accessible" was meant to be the "modern C# features".

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: > We are considering using the Mono runtime for the mobile platform.

Is "considering" the right word here? AFAIK, that's the only runtime that supports mobile platforms at the moment. NativeAOT might support them in the future.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: I wanted to make users more hesitant to opening Godot 3 projects with Godot 4, I feel like these changes makes it sound less scary.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
raulsntos: Should it be "into more way**s**"?

YuriSizov: Good spot!

collections/_article/whats-new-in-csharp-for-godot-4-0.md
YuriSizov: Yep, fair, I'll change it.

collections/_article/whats-new-in-csharp-for-godot-4-0.md
YuriSizov: (resolved on chat)

collections/_article/whats-new-in-csharp-for-godot-4-0.md
YuriSizov: (resolved on chat)

collections/_article/whats-new-in-csharp-for-godot-4-0.md
YuriSizov: Added a note about GodotSharp to hook it back to the original point. (also resolved on chat).

YuriSizov:
Thanks again for making this article!

PR #579, opened 2023-02-27
Update website for the imminent Godot 4 release

YuriSizov:
~This is still WIP.~ I've updated the Home page and the Features page with support for two stable releases at the same time, and with a new copy on top of the last Features overhaul. I'm still working on the Downloads page, but I wanted to publish this as a PR to start getting feedback (and possibly reviews on the copy).

chrome_2023-02-27_18-24-44
chrome_2023-02-27_18-24-04
chrome_2023-02-27_18-24-20
chrome_2023-02-27_18-24-29


YuriSizov:
Here are download pages. Godot 4 uses a new hero image to further differentiate it from Godot 3 (and to use a relevant demo project). Instructions have been adjusted as well, including removal of some alternative sources for Godot for macOS and Windows, since those are unlikely to be available on the release day.

Also slightly changed styling of the page, namely the platform selector section.

chrome_2023-02-28_00-10-04
chrome_2023-02-28_00-10-14
chrome_2023-02-28_00-10-26
chrome_2023-02-28_00-10-39

gomaproi:
There could be a little link that displays a popup explaining which version they should pick, a beginner that's picking up the engine for the first time probably can't tell the difference between the two versions. Maybe It could just be a bullet list briefly explaining that 4.0 has the latest features, but the LTS version is "battle-tested" or something like that?

YuriSizov:
That's a good idea, but I think we would have to handle it at a later date and throughout the website.

coppolaemilio (approved):

PR #589, opened 2023-03-03
Fix case of macOS in download layout

JohnVeness:


_layouts/download.html
```diff
@@ -175,8 +175,8 @@ <h2>Supported platforms</h2>
 				<span>Linux</span>
 			</a>
 			<a href="/download/macos" class="tab title-font {% if page.platform == 'MacOS' %} active {% endif %}">
```
bruvzg:
```suggestion
			<a href="/download/macos" class="tab title-font {% if page.platform == 'macOS' %} active {% endif %}">
```
This one probably should be changed as well.

JohnVeness: I saw that, but it's not user-visible text. Also, I'm not sure what page.platform returns, and whether the comparison is case-sensitive. If someone knows for sure, feel free to change!

_layouts/download.html
```diff
@@ -175,8 +175,8 @@ <h2>Supported platforms</h2>
...
-				<img width="24" height="24" src="/assets/icons/macos.svg" title="MacOS" alt="MacOS" />
-				<span>MacOS</span>
+				<img width="24" height="24" src="/assets/icons/macos.svg" title="macOS" alt="macOS" />
+				<span>macOS</span>
```
YuriSizov, line 179: Same should be done to the 3.x downloads page, right?

_layouts/download.html
YuriSizov: Actually, it already returns `macOS`, so this indeed needs to be changed. On both pages as well.

_layouts/download.html
bruvzg: ~This file seems to be used for both, probably no extra changes required~ There's `download-3.html` to change as well (it's `macOS` since 2016, so relevant for all versions).

YuriSizov: Yeah, we have dedicated layouts so it's easier to adjust incompatible differences :)

YuriSizov (approved):

YuriSizov:
Thanks!

PR #597, opened 2023-03-10
Add zenva and move other sponsors

coppolaemilio:
Home page:
image

Donate page:
image


pages/donate.html
```diff
@@ -496,8 +511,24 @@ <h3>Corporate Silver</h3>
 			</a>
 			{% endfor %}
 		</div>
-		<p class="for-more" style="margin-top: 50px; text-align: left; opacity: 0.5;">For corporate sponsorships please write to <a href="mailto:contact@godot.foundation">contact@godot.foundation</a>.</p>
+		<p class="for-more" style="margin-top: 50px; text-align: left; opacity: 0.8;">For corporate sponsorships please write to <a href="mailto:contact@godot.foundation">contact@godot.foundation</a>.</p>
 	</div>
 </section>
 
+<script>
+	document.addEventListener('DOMContentLoaded', () => {
```
YuriSizov, line 519: This seems like something we can do during generation.

coppolaemilio: I thought about it, but I wasn't able to get liquid to catch it. If you find a way for jekyll to check if one of those entries are empty and you can replace it please let me know 🙏 

The only solution I could find was setting the value to "blank" or something like that, but I prefer this since you can remove the section all together.

YuriSizov: No worries, we can merge it as is, and I can look into it later.

PR #599, opened 2023-03-14
Add Epic Games Store link to the downloads page

YuriSizov:
We have finally arrived! (An announcement will be published later this week)

This link is only available on the Godot 4 version of the page, so to account for that (and its general appearance) I had to adjust the layout a bit. I went with EGS and not the full name, because frankly it brings a lot of disbalance to this section whichever way I would put it. "Epic Games Store" is way longer than "itch.io" or "Steam", so it just doesn't work. Hope that's okay. The button's tooltip says the full name of the platform, by the way.

<details>
<summary>Godot 4 pictures!</summary>

chrome_2023-03-14_18-34-38
chrome_2023-03-14_18-34-48
chrome_2023-03-14_18-37-04

</details>

<details>
<summary>Godot 3 pictures!</summary>

chrome_2023-03-14_18-34-56
chrome_2023-03-14_18-35-05
chrome_2023-03-14_18-37-56

</details>

This PR also optimizes some of the existing SVG icons on the page.


_layouts/download.html
```diff
@@ -149,20 +149,25 @@ <h3>Instructions</h3>
 						</div>
 						<div class="digital-store-name">Steam</div>
 					</a>
+					<a class="btn btn-digital-store" href="https://store.epicgames.com/en-US/p/godot-engine" title="Godot Engine on Epic Games Store">
```
coppolaemilio: The link also works without the `/en-US/` part, so it might be better to let them do the redirect to the user locale?

I'll ask about the link just in case.

YuriSizov: Yeah, if it works universally, it's better to remove it.

```suggestion
					<a class="btn btn-digital-store" href="https://store.epicgames.com/p/godot-engine" title="Godot Engine on Epic Games Store">
```

coppolaemilio: They confirmed that `https://store.epicgames.com/p/godot-engine` is the correct link to share 👍

coppolaemilio (approved):

PR #618, opened 2023-04-12
Add Direct3D 12 shaders technical discussion blog post

RandomShaper:


collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
+---
+title: "Direct3D 12: Shaders with a twist"
+excerpt: "Godot is about to get a Direct3D 12 rendering driver. This is a discussion on the exotic approaches taken to face some of the challenges posed in the area of shaders."
+categories: ["progress-report"]
+author: Pedro J. Estébanez
+image: /storage/blog/covers/d3d12-shaders-with-a-twist.jpg.jpg
```
Calinou:
```suggestion
image: /storage/blog/covers/d3d12-shaders-with-a-twist.jpg
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+date: 2023-04-12 20:00:00
+---
+
+This is Pedro, a.k.a. [RandomShaper](https://twitter.com/RandomPedroJ) in the Godot community.
+
+You may already know what I have been implementing a **Direct3D 12 rendering driver** for Godot. Since [the announcement](https://w4games.com/2022/08/12/w4-games-donates-direct3d-12-support-to-godot/) that W4 Games donated my time working on that to the project, the code has received an amount of maintenance. In fact, [the pull request](https://github.com/godotengine/godot/pull/70315) hasn't still been merged since there has been a continuous flow of improvements.
```
Calinou:
```suggestion
You may already know that I have been implementing a **Direct3D 12 rendering driver** for Godot. Since [the announcement](https://w4games.com/2022/08/12/w4-games-donates-direct3d-12-support-to-godot/) that W4 Games donated my time working on that to the project, the code has received a lot of maintenance. In fact, [the pull request](https://github.com/godotengine/godot/pull/70315) hasn't still been merged since there has been a continuous flow of improvements.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+In this article I'd like to discuss specifically how I faced some of the challenges posed by different aspects of making shaders work, which was by far the most technically interesting area. Be warned that this piece is very dense in technical information.
```
Calinou:
```suggestion
In this article, I'd like to discuss specifically how I faced some of the challenges posed by different aspects of making shaders work. This was by far the most technically interesting area. Be warned that this piece is dense in technical information.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Translation of GLSL/SPIR-V to DXIL
+
+Godot rendering architecture is built with the industry standard Vulkan graphics API in mind. And that is great. However, once you want to plug new rendering APIs into the engine, you'll find that some elements need to be wired in a manner that they work with the alternative API (Direct3D 12 in this case.)
```
Calinou:
```suggestion
Godot's rendering architecture is built with the industry standard Vulkan graphics API in mind. And that's great! However, once you want to plug new rendering APIs into the engine, you'll find that some elements need to be wired in a manner that they work with the alternative API (Direct3D 12 in this case).
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+Shaders in Godot are written in its custom, very **GLSL**-like language. (The builtin shaders are already GLSL in the source code.) Shader code in this language is compiled by the engine (thanks to **glslang**) into SPIR-V. which is a well-defined format that stores the binary representation of a shader.
```
Calinou:
```suggestion
Shaders in Godot are written in its custom, very **GLSL**-like language. (The builtin shaders are already GLSL in the source code.) Shader code in this language is compiled by the engine (thanks to **glslang**) into SPIR-V, which is a well-defined format that stores the binary representation of a shader.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+SPIR-V and Vulkan play nice. However, Direct3D 12 features its own binary shader format: **DXIL**. This format is structured as a container with its most important contained chunks being shader metadata and the compiled shader code in the form of **LLVM bitcode** (yes, the same LLVM/Clang uses for an intermediate representation of the compiled code.) As a consequence, this rendering driver would need a way to generate DXIL from either the GLSL or the SPIR-V representations of the Godot shaders.
```
Calinou:
```suggestion
SPIR-V and Vulkan play nice. However, Direct3D 12 features its own binary shader format: **DXIL**. This format is structured as a container with its most important contained chunks being shader metadata and the compiled shader code in the form of **LLVM bitcode** (yes, the same LLVM/Clang uses for an intermediate representation of the compiled code). As a consequence, this rendering driver would need a way to generate DXIL from either the GLSL or the SPIR-V representations of the Godot shaders.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+### First round: SPIRV-Cross
+
+After assessing the available third-party libraries and open source projects capable of doing that (and license-compatible), I found **SPIRV-Cross** was the most sensible choice: it would translate GLSL to HLSL (the high level counterpart of GLSL in the DirectX world). From that point, the D3D12 renderer would use the standard **DirectX Shader Compiler** library to compile HLSL into DXIL.
```
Calinou:
```suggestion
After assessing the available third-party libraries and open source projects capable of doing that (and license-compatible), I found **SPIRV-Cross** was the most sensible choice: it would translate GLSL to HLSL (the high-level counterpart of GLSL in the DirectX world). From that point, the D3D12 renderer would use the standard **DirectX Shader Compiler** library to compile HLSL into DXIL.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+That worked quite well, but wasn't without big headaches about writing the code that should match the resource bindings in the Vulkan-targeted shaders to those in the DXIL ones. I won't delve into details on that here. Just know it required an insane amount of code and relying on certain SPIRV-Cross specifics. For now, the takeaway is that it was a hell of an approach, but worked in practice! (If you're curious, in the [first version of the pull request](https://github.com/godotengine/godot/pull/64304), you can see for yourself what I'm talking about.) This process also allowed me to make a few little contributions to SPIRV-Cross itself and learn its internals, which is always a good thing.
+
+### Second round: Via Mesa's NIR
+
+Months went by and I verified that the Microsoft was making great progress with **Dozen**, the Vulkan-to-D3D12 translation layer found in **Mesa**.
```
Calinou:
```suggestion
Months went by and I verified that Microsoft was making great progress with **Dozen**, the Vulkan-to-D3D12 translation layer found in **Mesa**.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+Mesa was already capable of translating SPIR-V into its own intermediate representation (**NIR**), but Dozen would have added the ability to translate NIR to DXIL. That looked very promising because it would be a binary-to-binary translation (i.e., SPIR-V to DXIL) without having to go all the way back to a high level language (GLSL/HLSL), which is slower and less convenient.
```
Calinou:
```suggestion
Mesa was already capable of translating SPIR-V into its own intermediate representation (**NIR**), but Dozen would have added the ability to translate NIR to DXIL. That looked very promising because it would be a binary-to-binary translation (i.e., SPIR-V to DXIL). This would avoid having to go all the way back to a high level-language (GLSL/HLSL), which is slower and less convenient.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+With some experimentation, I could make a shell script (included in the source code of the D3D12 renderer) that extracts from the Mesa source tree the relevant pieces of code to make the SPIR-V to DXIL translation possible.
+
+I wouldn't say that this integration was a piece of cake, but had major benefits in comparison to the SPIRV-Cross based approach. For instance, the bindings matching issue was easily solved by a Godot-specific patch to the Mesa code by which the translation process gets some callbacks. With that mechanism, the D3D12 renderer is told about the original set-binding pair of every resource in the SPIR-V shader and then it trivially maps them to some register indices in the DXIL world with some simple arithmetic:
```
Calinou:
```suggestion
I wouldn't say that this integration was a piece of cake, but it had major benefits in comparison to the SPIRV-Cross-based approach. For instance, the bindings matching issue was easily solved by a Godot-specific patch to the Mesa code by which the translation process gets some callbacks. With that mechanism, the D3D12 renderer is told about the original set-binding pair of every resource in the SPIR-V shader and then it trivially maps them to some register indices in the DXIL world with some simple arithmetic:
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+> DXIL register = SPIR-V set * 100000000 + binding * 100000
+
+A couple of notes on this:
+- We can make such a liberal use of DXIL registers since they are logical, just a way of tagging the bindings. (In past versions that was not the case, as far as I can tell.)
+- Regarding the last multiplication, if we understand Vulkan numbered bindings as "slots" and so we do with Direct3D register indices, in principle there would be no reason for it. However, Vulkan slots can hold entire arrays, whereas D3D ones need one slot per array element. Therefore, with that multiplication we are reserving a big range of D3D bindings to account for however big a incoming Vulkan array may be.
```
Calinou:
```suggestion
A couple of notes on this:

- We can make such a liberal use of DXIL registers since they are logical, just a way of tagging the bindings. (In past versions that was not the case, as far as I can tell.)
- Regarding the last multiplication, if we understand Vulkan numbered bindings as "slots" and so we do with Direct3D register indices, in principle there would be no reason for it. However, Vulkan slots can hold entire arrays, whereas D3D ones need one slot per array element. Therefore, with that multiplication, we are reserving a big range of D3D bindings to account for however big an incoming Vulkan array may be.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+In conclusion, the NIR way was kind of uncharted territory, but turned out to be much more convenient. I had to make a few PRs to Mesa to add or fix little pieces of translation logic to this part of Dozen, though. But again, that's a great learning experience.
+
+## Specialization constants (creatively working around the lack thereof in Direct3D)
+
+![SCs in DXIL](/storage/blog/d3d12/scs-in-dxil.jpg)
+
+For the first version of the PR (the SPIRV-Cross + DirectX Shader Compiler way) I published [a Twitter thread](https://twitter.com/RandomPedroJ/status/1532725156623286272?s=20) explaining how I approached this problem. Some parts are still true in the NIR-based PR. In any case, in this writeup I'm doing the full updated explanation of how it works now, so there's no need to read the other one.)
```
Calinou: Remove extraneous closing parenthesis:

```suggestion
For the first version of the PR (the SPIRV-Cross + DirectX Shader Compiler way) I published [a Twitter thread](https://twitter.com/RandomPedroJ/status/1532725156623286272?s=20) explaining how I approached this problem. Some parts are still true in the NIR-based PR. In any case, in this writeup I'm doing the full updated explanation of how it works now, so there's no need to read the other one.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+Vulkan features specialization constants (SCs from now on), which are some sort of compile-time constants in terms of optimization of the final shader code that runs on the GPU (branch pruning, informed loop unrolling, etc.), yet much more flexible, in that they can be applied at runtime to create the graphics/compute pipelines with **shaders already compiled to SPIR-V**.
```
Calinou:
```suggestion
Vulkan features specialization constants (SCs from now on), which are some sort of compile-time constants in terms of optimization of the final shader code that runs on the GPU (branch pruning, informed loop unrolling, etc.). However, these are much more flexible, as they can be applied at runtime to create the graphics/compute pipelines with **shaders already compiled to SPIR-V**.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+In Direct3D the closest is actual constants or preprocessor definitions in the HLSL code. That means that, in order to create multiple pipelines corresponding to different variants of the same shader, you have to compile multiple versions of the HLSL code with different values set each time at the source code level.
```
Calinou:
```suggestion
In Direct3D, the closest is actual constants or preprocessor definitions in the HLSL code. That means that, in order to create multiple pipelines corresponding to different variants of the same shader, you have to compile multiple versions of the HLSL code with different values set each time at the source code level.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+What I needed is that those operations are not performed offline, but be promoted to true dynamic operations, **as if the original GLSL had used non-constant variables instead of SCs**, and therefore not optimizable by now.
+
+Mesa's SPIR-V to NIR code is patched in the Godot D3D12 renderer to do precisely that. The values of the SCs are set to some **carefully chosen magic sentinel value plus the SC id**, so in a later step we can tell if a literal value corresponds to an SC, and which one.
+
+Some extra NIR operations are added depending on the type of the SC. For instance, for a `float`, there's some bit casting so the base value can be treated as a plain 32-bit integer but the shader eventually sees it as a floating point value and thus does the right thing with it once patched.
```
Calinou:
```suggestion
Some extra NIR operations are added depending on the type of the SC. For instance, for a `float`, there's some bit casting so the base value can be treated as a plain 32-bit integer but the shader eventually sees it as a floating-point value and thus does the right thing with it once patched.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+I patched the code that emits constants to the DXIL stream so that it recognizes cases of SCs if they match the sentinel value mask. The rest of the bits are the SC id. On each occurrence, the corresponding callback is leveraged to provide the information to the Godot D3D12 driver. (It's a bit more complicated than that, since the offset in the bitcode has later to be adjusted to match where the `DXBC` chunk ends up in the whole DXIL blob.)
+
+### 4. Creating pipelines with patched DXIL
+
+The Direct3D rendering driver is now able to apply whatever specialization constant values are needed to the already compiled shader, which was our initial goal.
+
+Now I have to mention that LLVM bitcode is an akward beast and that it's not trivial to patch a miserable integer due to variable bit-rate encoding. (Speaking of that, the former approach I implemented had a limit on the number of bits that could be patched, which imposed big limitations in the usable range of integers and imposed an elephant-sized epsilon for floating point ones; those have been happily lifted in the current one.)
```
Calinou:
```suggestion
Now I have to mention that LLVM bitcode is an akward beast and that it's not trivial to patch a miserable integer due to variable bit-rate encoding. (Speaking of that, the former approach I implemented had a limit on the number of bits that could be patched. This imposed big limitations in the usable range of integers and imposed an elephant-sized epsilon for floating point ones; those have been happily lifted in the current one.)
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+This is the comparison of the assembly code produced by the same shader, only with different SC values, which allowed in the second case to optimize out a multiplication:
+![Sample ISA code A](/storage/blog/d3d12/isa-a.png)
+![Sample ISA code B](/storage/blog/d3d12/isa-b.png)
```
Calinou:
```suggestion
This is the comparison of the assembly code produced by the same shader, only with different SC values, which allowed in the second case to optimize out a multiplication:

![Sample ISA code A](/storage/blog/d3d12/isa-a.png)

![Sample ISA code B](/storage/blog/d3d12/isa-b.png)
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+Microsoft may very well end up adding SCs natively to Direct3D. Nonetheless, in the current state of things this hack/technique/magic is **very close to what real SCs would be in terms of performance**. The ISA assembly takes the patched values into account so it can run its own informed optimization passes and so provide a late form of what we were making an effort to prevent at earlier stages. Maybe the biggest downside is that the DXIL blob has to be validated-signed every time it's patched. which takes time (not much, though) and forces us to include the `dxil.dll` file with the binary distribution of the engine.
```
Calinou:
```suggestion
Microsoft may very well end up adding SCs natively to Direct3D. Nonetheless, in the current state of things this hack/technique/magic is **very close to what real SCs would be in terms of performance**. The ISA assembly takes the patched values into account so it can run its own informed optimization passes. This way, it can provide a late form of what we were making an effort to prevent at earlier stages. Maybe the biggest downside is that the DXIL blob has to be validated-signed every time it's patched. This takes some time and forces us to include the `dxil.dll` file with the binary distribution of the engine.
```

Calinou (changes requested):
Great article! I made some grammar/stylistic changes to try and make the article flow better.

collections/_article/d3d12-shaders-with-a-twist.md
Ansraer:
```suggestion
You may already know what I have been implementing a **Direct3D 12 rendering driver** for Godot. Since [the announcement](https://w4games.com/2022/08/12/w4-games-donates-direct3d-12-support-to-godot/) that W4 Games donated my time working on that to the project, the code has received an amount of maintenance. In fact, [the pull request](https://github.com/godotengine/godot/pull/70315) still hasn't been merged since there has been a continuous flow of improvements.
```

collections/_article/d3d12-shaders-with-a-twist.md
Ansraer: I am struggling with "(yes, the same LLVM/Clang uses for an intermediate representation of the compiled code)."
Should either be "[...]the same LLVM/Clang used for[...]" or "[...]the same bitcode LLVM/Clang uses for an intermediate representation of compiled code[...]".

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+A couple of notes on this:
+- We can make such a liberal use of DXIL registers since they are logical, just a way of tagging the bindings. (In past versions that was not the case, as far as I can tell.)
+- Regarding the last multiplication, if we understand Vulkan numbered bindings as "slots" and so we do with Direct3D register indices, in principle there would be no reason for it. However, Vulkan slots can hold entire arrays, whereas D3D ones need one slot per array element. Therefore, with that multiplication we are reserving a big range of D3D bindings to account for however big a incoming Vulkan array may be.
+
+About mapping the Vulkan resource types (image, buffer, etc.) to their D3D12 counterparts (SRV, UAV, etc.), there's no 1:1 mapping. It depends on extra attributes that decorate each binding, and even on actual usage! With the SPIRV-Cross approach, the Godot D3D12 renderer needed to do reflection on the DXIL shader to find out (reflection was used for the bindings, too.) In contrast, with the NIR way, the callbacks mentioned earlier can also get information about which kind of Direct3D resource each binding is being translated to, on the go.
```
Ansraer:
```suggestion
When mapping the Vulkan resource types (image, buffer, etc.) to their D3D12 counterparts (SRV, UAV, etc.), there's no 1:1 mapping. It depends on extra attributes that decorate each binding, and even on actual usage! With the SPIRV-Cross approach, the Godot D3D12 renderer needed to do reflection on the DXIL shader to gather additional information (reflection was used for the bindings, too). In contrast, with the NIR-based approach, the callbacks mentioned earlier can also get information about which kind of Direct3D resource each binding is being translated to, on the go.
```

collections/_article/d3d12-shaders-with-a-twist.md
```diff
@@ -0,0 +1,112 @@
...
+
+In Direct3D the closest is actual constants or preprocessor definitions in the HLSL code. That means that, in order to create multiple pipelines corresponding to different variants of the same shader, you have to compile multiple versions of the HLSL code with different values set each time at the source code level.
+
+SPIRV-Cross does some work of assisting with that, making that process of patching the source code a bit more ergonomic, but the issue of having to re-compile the source is still present. Moreover, the PR in its current, definitive form is using an approach where the source code is not used at all (neither in GLSL nor translated to HLSL.) Therefore, some way of applying different values to SCs to shaders compiled to DXIL is a must. (In the SPIRV-Cross way it was a nice-to-have feature, but patching HLSL and recompiling, even if very slow and forcing to store the source code in the shader cache, would have done the trick.)
```
Ansraer:
```suggestion
SPIRV-Cross does some work of assisting with that, making the process of patching the source code a bit more ergonomic, but the issue of having to re-compile the source is still present. Moreover, the PR in its current, definitive form is using an approach where the source code is not used at all (neither in GLSL nor translated to HLSL). Therefore, some way of applying different values to SCs for shaders compiled to DXIL is a must. (In the SPIRV-Cross way it was a nice-to-have feature, but patching HLSL and recompiling, even if very slow and forcing to store the source code in the shader cache, would have done the trick.)
```

Ansraer (commented):
I am not a native speaker, so take these suggestions with a grain of salt.

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: I was overapplying the period-inside-punctuation-segment rule. I think it's the right way to go with double quotes at least. But, yes, not applicable to parentheses, which seem to behave as they do in Spanish in relation to periods.

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: I've changed it to, "The closest in Direct3D is [...]."

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: You helping me avoiding overly long sentences once more. 😀

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: I see the point of the first line break. But in the other case what I actually want is both images side-by-side (in fact, I've replaced the line break I added there with a whitespace). They are very thin and will likely fit the screen width everywhere (they second one will just be wrapped into the next line otherwise, won't it?). Just let me know if by doing this I'm missing some rule we have, some layout limitation, or whatever else.

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: Accepted everything, but edited the last change to read as, "This takes some extra time [...]".

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: I think there's a rule to put adverbs between the two words of a compound verb. CC @Calinou

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: I'll change it to, "[...] the same kind of bitcode [...]".

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: All applied, with some changes:
- First sentence reworded as, "The operation of mapping the Vulkan resource types (image, buffer, etc.) to their D3D12 counterparts (SRV, UAV, etc.) is not trivial. The exact resource type depends on [...]".
- The middle subsentence reworded as, "[...] do reflection on the DXIL shader to find out the D3D12 types the Vulkan ones ended up being mapped to".

RandomShaper:
Thank you for the review! I'm pushing with most suggestions applied (slightly edited in the cases I've commented).

RandomShaper:
I've pushed again, changing the title (and filename).

collections/_article/d3d12-shaders-with-a-twist.md
YuriSizov: We can add custom markup to make them appear side by side on wide enough screens and one under the other on mobile. Markdown allows to inject HTML.

collections/_article/d3d12-shaders-with-a-twist.md
Calinou: "still hasn't been merged" sounds more natural to me.

collections/_article/d3d12-shaders-with-a-twist.md
Calinou: @RandomShaper You forgot to apply the 8 suggestions that were folded by GitHub :slightly_smiling_face:

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: Applied, but I put the hyphen in "high-level," not in "level-language," as I think your intent was.

collections/_article/d3d12-shaders-with-a-twist.md
RandomShaper: Won't that just happen automatically? In other words, won't image-space-image be rendered as a set of inline elements?

RandomShaper:
Fixed missing pieces of feedback. The only remaining thing is about the layout of images.

collections/_article/d3d12-shaders-with-a-twist.md
YuriSizov: No, it won't happen automatically.

YuriSizov:
> The only remaining thing is about the layout of images.

We also need a new cover image, up to our current standards :) @coppolaemilio is on the case though.

RandomShaper:
> We also need a new cover image, up to our current standards.

I worked on that for weeks!

Now seriously, can you point me to the guidelines?

YuriSizov:
We don't have any formal guidelines (not even sure what they'd look like), but you can get a feel for what we aim at on the last couple of pages of blog posts (e.g. https://godotengine.org/blog/2/). We either have a blog post with a nice thematic background and a standard overlay, or we have a clean looking custom image without text.

collections/_article/d3d12-shaders-with-a-twist.md
clayjohn: This is how it ends up looking 
image

clayjohn: isa-joined

Its probably easiest to just manually join them like this

YuriSizov: Adding custom markup is easy, and will make sure it works nice on mobile. If you merge them together, it won't work nice on mobile at all. I'll add a suggestion with the markup tomorrow, we should have it in some earlier posts, so it shouldn't take long.

RandomShaper: I have a big temptation to use a table (like in the good old times). I can use some divs with flex layout, but I have the impression that tables will survive RSS readers better, despite how non-semantic that would be.

RandomShaper: I didn't realize tables wouldn't work if I want wrapping. I've pushed with some flex-based styling. Would that work?

clayjohn: Looks like no

Screenshot from 2023-04-13 14-41-16

RandomShaper: Thanks for testing, Clay. Could you do one more? If it doesn't work this time, I'll start doing it myself.

clayjohn: Looks good now!

Screenshot from 2023-04-13 15-13-15

Also tested with a tall aspect ratio
Screenshot from 2023-04-13 15-13-32

YuriSizov: Note that we don't put full text in our RSS feed :)

Ansraer: Maybe you could also center the images?

RandomShaper: Done!

coppolaemilio (approved):

coppolaemilio:
Thank you :)

PR #624, opened 2023-04-19
Reduce vertical space in Blog

coppolaemilio:
This PR moves the search bar to the hero section and then makes the categories and the title in one line. This reduces a big amount of vertical space so it allows for more blog posts to be visible.

## Currently:
Screen Shot 2023-04-19 at 21 29 18

## With this PR:
Screen Shot 2023-04-19 at 21 32 08

### DIFF
--- a/_layouts/blog.html
+++ b/_layouts/blog.html
@@ -27,30 +27,67 @@ collection: article
 			grid-template-columns: 1fr 1fr 1fr;
 		}
 	}
+
+	.head .main {
+		display: grid;
+		gap: 30px;
+		padding: 20px 0 39px;
+	}
+	.head .main > :last-child,  
+	.head .main > :first-child {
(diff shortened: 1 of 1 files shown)


YuriSizov:
This is really nice! Do you think that the search bar fits our design? I'm kind of underwhelmed by it, and with you moving it to the header section it's even more meh IMO.

coppolaemilio:
Thanks!
I don't really like how the entire hero looks like, but felt like this was the easiest change to improve it 😅 
Maybe starting a new section from scratch could help, but would reserve it for a future PR.

akien-mga:
We should at least fix the contrast for the search bar on the light theme:
image

And from the above screenshot, the contrast on dark theme isn't the best either (the search button is good, but the dark grey background on dark blue hero doesn't stand out).

PR #625, opened 2023-04-20
Fix a typo introduced in the release management post right before merging

clayjohn:


akien-mga:
To be fair, it works equally with both, especially since it says "cycle" (while 3.x was multiple cycles) ;)

clayjohn:
> To be fair, it works equally with both, especially since it says "cycle" (while 3.x was multiple cycles) ;)

In that sentence yes, but in the points below we refer to "releases" and feature freeze "periods" as plural which doesn't make sense if we are only talking about 4.0

PR #628, opened 2023-04-21
Rework community events using data collections

YuriSizov:
It should be much easier to update this page going forward, to add new events and move past events to the appropriate section. It should also be possible to generate dedicated event pages in the future, so they can be referenced or printed out.

Visually and functionally the page should remain the same. Two minor changes include a subsection title before the event details and a rename of the map link. Also fixes some markup issues and typos.

-----

I also added templates for various collection items, so we can have a good view on their format, parameters, and just create new docs with more ease. This actually exposed that showcase items used `page.xbox` for two separate purposes, as a bool and as a store link. I fix this "shadowing" issue, but we may want to scope those flags in the future. Also added support for PlayStation store links.


_layouts/showcase-item.html
```diff
@@ -141,12 +141,18 @@ <h3 style="opacity: 0.75; margin-top: 0.5rem">
 				</a>
 				{% endif %}
 
-				{% if page.xbox %}
-				<a href="{{ page.xbox }}" class="btn" style="margin: 0 0.5rem; white-space: nowrap" target="_blank">
+				{% if page.microsoft %}
```
coppolaemilio, line 144: Any reason for the rename? Microsoft themselves are using xbox as the brand for the gaming side of things, and the url used in Primal Light is also an xbox URL `https://www.xbox.com/en-US/games/store/primal-light/9N2XZFRWG6R8`

coppolaemilio:
Found this string to be a bit weird:
image

Is it a debugging print?

_layouts/showcase-item.html
YuriSizov: Yes, the reason is the name collision:

> I also added templates for various collection items, so we can have a good view on their format, parameters, and just create new docs with more ease. This actually exposed that showcase items used `page.xbox` for two separate purposes, as a bool and as a store link. I fix this "shadowing" issue, but we may want to scope those flags in the future. Also added support for PlayStation store links.

I think a proper solution would be to use objects and group those properties, but for now a simple renaming fix should be good enough.

YuriSizov:
> Is it a debugging print?

Oops, yes. Fixed!

coppolaemilio:
Thanks!

PR #629, opened 2023-04-21
Create next and previous buttons below the articles

### LINKED ISSUE #419: Add links to next and previous news articles

a2937:
I created a very simple div in the article layout right below the article. If there is previous url, it adds a hyperlink to the previous article. And if there is another article after that one, it adds a link to it. 

Closes #419


Calinou (changes requested):
Tested locally (rebased on top of `master` https://github.com/godotengine/godot-website/commit/f19d8fd197a939e40208a600a3cc53cb0e81bb22), it works as expected. Sorry for the delay in reviewing.

The navigation is functional, but there's not really much design going around the previous/next links, so they look out of place:

image

___

image

I suggest adding padding around the `previous-next` class, and perhaps give it a background color similar to the article as well.

Also, it might be a good idea to repeat this previous-next block at the top of the article for easier navigation (i.e. the controls should be both at the top and bottom of the article).

_layouts/article.html
```diff
@@ -102,6 +102,11 @@
 		align-items: center;
 	}
 
+
+	 .next {
+    float: right;
+  }
```
Calinou: Use tab indentation only:

```suggestion
	.next {
		float: right;
	}
```

_layouts/article.html
```diff
@@ -145,6 +150,7 @@
 
 <link rel="stylesheet" href="/assets/css/highlight.obsidian.min.css">
 
+
```
Calinou: Use a single blank line only:

```suggestion

```

_layouts/article.html
```diff
@@ -195,6 +201,14 @@ <h1>
 			</div>
 		</div>
 	</article>
+		<div class="Previous-next">
```
Calinou: Use lowercase class name:

```suggestion
		<div class="previous-next">
```

coppolaemilio:
I've made some design changes, but I'm not a fan of the feature. It seems useful when you first hear about it, but when do you want to navigate this way?

I doubt anyone would actually use it to browse articles, and it might be much much better to just have a link back to the full blog, where you have the search bar, the categories, etc...

I noticed this after styling it locally and testing it out. It was never something that I would use because it isn't even context aware. If it was only jumping from release to release, or linking articles that make sense I would understand, but chronologically doesn't. 

So, if no one is super crazy about this feature I would advice against it, but if you all think it should be added here are the changes I made. I can push to this PR if you like them:

## Desktop
image
image

## Mobile
image
image

Pinging @JohnVeness

JohnVeness:
Hi, thanks for working on this.

It's probably not a feature that someone would use if they regularly read articles as they come out, or if they just read articles they are particularly interested in. For me, I find it useful if I want to catch up on a project that is new to me, or which I haven't checked on in some months. In those cases, I click on the oldest new article, read it, click on the next article link at the bottom, and so on. This is more comfortable (and usually quicker, depending on site speed) than navigating back and finding the next article manually.

If there are no overriding negatives about this feature, I would still vote for it being there, even if its use is minimal.

coppolaemilio:
Chatted a bit with @Calinou and it seems like yes, people do use it, so I guess I'll merge 😅 

Thanks!

PR #640, opened 2023-05-16
Add a download page for preview builds

YuriSizov:
This PR adds a dedicated download page for all on-going version previews (snapshots, betas, RCs). This allows us to make the builds "official" ahead of (or instead of) publishing complete release notes. This new page is not persistent, though, and just like normal download pages it updates with every change to the version list.

To make this work (and to lay some ground work for the future "All version downloads" page to replace TuxFamily) I made everything data-driven, using `versions.yml` as well as a couple of other data files. The setup is rather complex, but content management should be trivial now, thanks to those files and to the custom plugin that generates correct URLs for any known version.

Existing download pages have been updated to use the new plugin and data files as well. So please test them for regressions.

-----

Here's how the new page looks like:

image

<details>
<summary>A video overview</summary>

https://github.com/godotengine/godot-website/assets/11782833/c28e9900-9464-46ef-a740-21de06322a6a

</details>

I also changed the bottom of the downloads page to rework the awkward GitHub blurb and to add a link to the new page.

chrome_2023-05-16_14-57-45

-----

One small change that may require follow-up tuning is related to the download names. The top buttons are unaffected, but the lists of downloads are. I had to standardize some descriptors, and as a result "Standard" is no longer a counterpart of ".NET". It's just a descriptor, and ".NET" is added on top of that.

So new download link texts read "Platform - Standard" and "Platform - .NET - Standard". This is probably not ideal, but it's the best option I have for now. Perhaps in the future this will stop being an issue once we merge the editors.

[Before](https://github.com/godotengine/godot-website/assets/11782833/ba76331e-258a-4059-819a-86c026161011)
[After](https://github.com/godotengine/godot-website/assets/11782833/b38cb43b-44ce-4449-99a2-5f1163c10d24)


_data/download_configs.yml
```diff
@@ -0,0 +1,55 @@
...
+        windows.32: mono_win32.zip
+        linux_server.headless.64: mono_linux_headless_64.zip
+        linux_server.64: mono_linux_server_64.zip
+      extras:
+        aar_library: mono.release.aar
+
+  2:
+  1:
+
+overrides:
```
coppolaemilio: Why are these values here but empty?

_includes/header.html
```diff
@@ -7,7 +7,7 @@
 				<img class="nav-logo dark-logo" src="/assets/logo_dark.svg" width="136" height="48" alt="Godot Engine">
 			</a>
 			<label for="nav_toggle_cb" id="nav_toggle_btn">
-				<img src="/assets/hamburger.svg" width="24" height="24" alt="">
+				<img src="/assets/icons/hamburger.svg" width="24" height="24" alt="">
```
coppolaemilio:
```suggestion
				<img src="/assets/icons/hamburger.svg" width="24" height="24" alt="Menu">
```
I guess we should add an alt text here

_layouts/download-3.html
```diff
@@ -335,7 +381,7 @@ <h3>Thank you!</h3>
 		</ul>
 
 		<div class="btn-back-to-downloads">
-			<img src="/assets/cross.svg" width="24" height="24" alt="">
+			<img src="/assets/icons/cross.svg" width="24" height="24" alt="">
```
coppolaemilio:
```suggestion
			<img src="/assets/icons/cross.svg" width="24" height="24" alt="close">
```

_layouts/download.html
```diff
@@ -330,7 +370,7 @@ <h3>Thank you!</h3>
 		</ul>
 
 		<div class="btn-back-to-downloads">
-			<img src="/assets/cross.svg" width="24" height="24" alt="">
+			<img src="/assets/icons/cross.svg" width="24" height="24" alt="">
```
coppolaemilio:
```suggestion
			<img src="/assets/icons/cross.svg" width="24" height="24" alt="Close">
```

coppolaemilio:
There are some visual aspects of it that I would like to address as well:

Cards like this one, we usually round the corners, here they are square and it clashes with the rest of the pages.
image

This section should be either centered or made into a two column layout, and I would remove the gap between it and the footer:
image

These two are a bit similar in terms of messaging. I would make the link that goes to the article to say something like "Read more". I also think that both links could be in the bottom section, like
`Read More`
`View complete changelog`
With something like a button style for the Read More, and just text for the changelog one.
image

_data/download_configs.yml
YuriSizov: 1 and 2 I marked down for completeness. I'll fill them out when I work on the "download older versions" page, where they would be relevant. I expect there to be exceptions, which is where overrides would be needed, but so far it's no-op. I can drop it, if you want (but 1 and 2 I'd keep, will definitely need them in the future).

YuriSizov:
> Cards like this one, we usually round the corners, here they are square and it clashes with the rest of the pages.

We only recently started to round corners, before it was all mostly squared. And it's still squared on the community and events pages. I actually followed the events page here. But I can change it, if you want. Which radius should I use, do you have an element in mind that I can reference?

> This section should be either centered or made into a two column layout

You mean I should also center the title? Just like on the features page? I thought about it, with "Need a stable version?" as a title probably.

YuriSizov:
Okay, I made some changes.

chrome_2023-05-16_22-14-02
chrome_2023-05-16_22-14-14
chrome_2023-05-16_22-14-22

coppolaemilio:
Thanks!

PR #651, opened 2023-06-08
4.1-beta1: Add known issue for Windows AMD OpenGL crash

akien-mga:


YuriSizov (approved):

YuriSizov:
Thanks, I should've added it from the get-go!

PR #653, opened 2023-06-13
Add Adam Scott to the production team

adamscott:
Just to say that I'm only using the alphabetical order, here!

(redo of #652, but with a branch name without emojis)


YuriSizov (approved):

YuriSizov:
> Just to say that I'm only using the alphabetical order, here!

That would be correct :)

PR #657, opened 2023-06-19
Generate the mirror list based on static version data

### LINKED ISSUE #656: Generate mirrorlist based on the version data file
See https://github.com/godotengine/godot/issues/78375.

We would need to add some additional information to each version record, namely its previous release flavors and available mirror sources (TF for everything, our primary repo for stable and our secondary repo for pre-releases, starting with 3.5/4.0). But otherwise everything is static about the mirrorlist and it should've been done soon after we moved to the static website.

YuriSizov:
Closes https://github.com/godotengine/godot-website/issues/656.

The implementation is not as straightforward because instead of relying on collections, which would be quite excessive, we generate pages ourselves using a singular data file. I've extended the `versions.yml` to contain some information about pre-releases for each version, which is going to be useful for a future download page, but is also needed to generate a mirror list JSON for each official build.

There is also an extra config that controls which hosts are added to the mix for each major-minor release code (assuming patch releases are never so drastic to change their host composition). This is only violated with 3.1, which for some reason is not uploaded to GitHub releases (https://github.com/godotengine/godot/releases/tag/3.1-stable) while 3.1.1 is. I still generate the mirror list as if it was uploaded for simplicity, assuming nobody is going to be affected by this tiny issue in a .0 zero release from several years ago.

I also generate mirror lists up to 1.0, even though you can't download 1.0 at all, and there is no template manager until a much later version. It's simply not worth it to exclude those versions given that we already have all the version data.

Other than that, it generates the same as existing files. Speaking of which, **when deploying make sure to remove them from hosting before pulling the changes**.


YuriSizov:
Oh, btw, 3.0 was the first version to have the mirrorlist, but it used a format that differs from any following release. This PR preserves that bit of logic:

https://godotengine.org/mirrorlist/3.0-stable.json vs https://godotengine.org/mirrorlist/3.1.stable.json

PR #664, opened 2023-07-04
Update the Teams page with current/active members

YuriSizov:
- Removed **EricEzaM, AndreaCatania, naithar, RPicster,** and **cbscribe** from their teams due to inactivity at the moment.
- Added **dsnopek** to the GDExtension team, as the de facto leader too.
- Added **paddy-exe** to the Tech art team.
- Removed **reduz** from the listing for the Production team (he is still a part of it, but not on the day-to-day basis).
- Removed **mhilbrunner** from the Website team and added **coppolaemilio** to it.
- Set _myself_ as the de facto leader of the Website team in place of **Calinou**.

The rest are just formatting changes.

### DIFF
--- a/pages/teams.html
+++ b/pages/teams.html
@@ -52,7 +52,8 @@ current_tab: "teams"
 		</p>
 
 		<p>
-			<span class="teams-subteam-leader">Juan Linietsky (<a href="https://github.com/reduz">@reduz</a>)</span>,
+			<span class="teams-subteam-leader">Juan Linietsky (<a
+					href="https://github.com/reduz">@reduz</a>)</span>,
 			George Marques (<a href="https://github.com/vnen">@vnen</a>),
 			HP van Braam (<a href="https://github.com/hpvb">@hpvb</a>),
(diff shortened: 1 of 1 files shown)


pages/teams.html
```diff
@@ -52,7 +52,8 @@ <h4 class="title" id="core">Core • <a href="https://chat.godotengine.org/chann
 		</p>
 
 		<p>
-			<span class="teams-subteam-leader">Juan Linietsky (<a href="https://github.com/reduz">@reduz</a>)</span>,
+			<span class="teams-subteam-leader">Juan Linietsky (<a
+					href="https://github.com/reduz">@reduz</a>)</span>,
```
akien-mga, line 56: Are those formatting changes really improving things?
I had tried to format it with one name per line.

akien-mga (approved):

pages/teams.html
YuriSizov: We had mixed formatting, even for the same person but in different teams. I just picked one so it's at least consistent.

PR #665, opened 2023-07-04
Add godot 4.1 release post

NathanLovato:
I addressed the comments left on the shared Google doc: https://docs.google.com/document/d/1qk5sjhRBvD4UIU45O1ebegOlICQaAtDbBPlJrUnrj0U/edit.

Feel free to make any other changes directly on top of this PR


YuriSizov:
Note for future me: We should probably convert all images here to webp (and also replace spaces in file names with underscores or dashes).

raulsntos:
Can we include a section listing all the break compat changes?[^1]

I've been tracking breaking changes in this [gist](https://gist.github.com/raulsntos/761ffb4afd56b4ca46cef50bf060bdcb). I last checked with RC 1, but I don't expect any breaking changes to have been introduced after the first RC.

[^1]: Perhaps something similar to https://learn.microsoft.com/en-us/dotnet/core/compatibility/7.0.

NathanLovato:
@raulsntos sure, these are compatibility breakages on C# specifically, right? Seeing the info, I think it'd be good to have a dedicated file or page, probably in the docs, with this table you linked to, and to include the link in the release post. If I understand correctly, it's for people who use C# in 4.0 and who are porting their project to 4.1? If so, this should you could have a paragraph like this:

> If you have a Godot 4.0 project that uses C#, if you convert it to Godot 4.1, please note that some properties and methods were changed in this version and will need to be updated manually. You can see every changed method and property in the official docs: [C# compatibility changes in Godot 4.1](#)

Also, it'd be good to mention right away if that won't happen again in the future. E..g if the backwards compatibility system in GDExtension or any future development will prevent compatibility breakages. Because otherwise people may perceive that this could happen with every new release and C# in Godot may get some more bad rap.

raulsntos:
> these are compatibility breakages on C# specifically, right?

No, these are changes to Core so it would affect every language. Different languages handle this differently, for example in GDScript optional parameters aren't a breaking change but in C# they are. Some of the changes in the list will break GDScript too, for example the navigation changes. GDExtension is also affected, but it's in beta so breakage is expected there.

> If I understand correctly, it's for people who use C# in 4.0 and who are porting their project to 4.1?

Yes, but the biggest issue is if they are using a library that was built against GodotSharp 4.0. In this case they can't fix it because the problem is not in their code, so all they can do is wait for the library to be updated or get rid of the library.

> Also, it'd be good to mention right away if that won't happen again in the future.

Of course, it's not our intention to break compat and I think we have tried to avoid it. I think it's a good idea to reaffirm our commitment to this. However, I don't think we should say that it won't happen again. For example, the navigation API was marked experimental in 4.0 so breaking compat was expected there.

> E..g if the backwards compatibility system in GDExtension or any future development will prevent compatibility breakages.

We have already added compat methods to C# to avoid breaking compat in 4.1 as much as possible. But some of the breaking changes were unavoidable, for example when the only thing that has changed in a method is the return type. In C# you can't have method overloads when the only thing that changes is the return type.

> Because otherwise people may perceive that this could happen with every new release and C# in Godot may get some more bad rap.

While breaking compat can be disruptive, I think most users will likely not mind that much as long as the breaking changes are communicated properly. Listing the APIs that changed as well as steps to fix the broken code would allow users to upgrade easily; otherwise, the upgrade process is likely going to be a frustrating experience and that will lead to unhappy users.

That is to say, I wish we could avoid breaking compat entirely in future versions. However, if it happens, giving the users the expectation that we will always include this information in the release article, so they can see what changed and prepare accordingly, sounds like a good thing to me.

> I think it'd be good to have a dedicated file or page, probably in the docs, with this table you linked to, and to include the link in the release post.

Sure, if we prefer to have it in the documentation I can do that.

NathanLovato:
@raulsntos Ok that's great information. To me, all that information would be good to have along with a table in a compatibility breakage docs page. There are a couple of reasons I recommend a docs page:

1. It's docs material, it's a technical reference of sorts.
2. You want to be able to easily link this in the community, and not have to link the release post and scroll to find the compat. breakages.
3. Users and the dev team will benefit from having all this info, both where compat breaks, but also reassurance that you're trying your best to avoid it and why sometimes you really can't avoid it. It would prevent potential frustration.
4. But that much info would be heavy in a release post and break the communication with info that will not be accessible or relevant to a fairly large number of users.

YuriSizov:
What we need to add is a Known Issues section, like we do in other release notes. We have a couple of candidates for it:

- ~Node manipulation optimizations change how node names are validated, which affects import among other things. Periods are no longer removed, they are instead replaced with an underscore. As a result, this broke import hints (like `-col`) from being parsed correctly on import. https://github.com/godotengine/godot/issues/78881~ We merged https://github.com/godotengine/godot/pull/79050 directly into stable, it should be fixed for now.

- The new API to define custom export configuration options can lose data on editor reload. There is a fix available and we're likely to patch it in 4.1.1 https://github.com/godotengine/godot/pull/79025

RedworkDE:
> * The new API to define custom export configuration options can lose data on editor reload. There is a fix available and we're likely to patch it in 4.1.1

More like it doesn't (usefully) work at all currently, except for the use by the builtin C# export plugin.

YuriSizov:
I've added a couple of changes on top of the PR. First one updates images and meta. I didn't end up converting PNGs to WebPs as files are already tiny and I'm afraid of introducing compression artifacts at this point. Specific changes are outlined in the commit description.

The second improves the excerpt and introduction, adds some structure to the start of highlights and adds the known issues section with the sole entry listed above.

Another review pass would be welcome, then we should be good to go.

coppolaemilio (approved):

raulsntos:
Sorry for being insistent, but I think it's important to mention the breaking changes in the release post itself. I made a PR to the documentation as discussed. Can we merge it and link to it in the release post? (needs to be reviewed though).
- https://github.com/godotengine/godot-docs/pull/7611.

YuriSizov:
@raulsntos We still need to work on that page, but in the meantime we'll add a link to the PR itself to the article.

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,178 @@
...
+
+![The code editor detached from the main editor window](/storage/blog/godot-4-1-is-here/detached-code-editor.png)
+
+Still, we took great care to prioritize the bugs you have encountered in 4.0. This update fixes over 900 issues that users have reported from using Godot 4 or while helping contributors test 4.1 with pre-release builds. The engine should feel more reliable overall. We will continue improving stability, performance, and workflows with every upcoming feature release of Godot 4.
+
+For most games and apps made with 4.0 it should be relatively safe to migrate to 4.1. We are [preparing a migration guide](https://github.com/godotengine/godot-docs/pull/7611) that outlines everything you need to pay attention to when migrating your project. Some incompatibilities are expected for C# and GDExtension users specifically, however we are working on making sure to avoid that in future releases. Don't forget to always make backups when moving versions, even minor. Better yet, prefer using a version control system, such as Git, and commit a version of your project before the migration.
```
clayjohn, line 18: I would be careful about saying we are preparing a migration guide. So far the work done is just listing breaking changes to the scripting interface. If we say we are creating a migration guide we should have a plan + someone responsible to prepare the guide

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,178 @@
...
+
+Besides financial support, you can also give back by: writing detailed bug reports, contributing to the code base, writing documentation, writing guides (for the docs or on your own website), and supporting others on the various [community platforms](https://docs.godotengine.org/en/latest/community/channels.html) by answering questions and providing helpful tips.
+
+Last but not least, making games with Godot and crediting the engine goes a long way to help make it more popular and convince more people to contribute and improve Godot for everyone. Remember, we are all in this together and Godot requires community support in every area to thrive.
+
+## Highlights
+
+This release was made possible thanks to submissions from over 300 contributors! We warmly thank you all for your work and dedication, including those who help us test the engine before the release and submit bug reports.
```
clayjohn:
```suggestion
This release was made possible thanks to submissions from over 300 contributors! We warmly thank you all for your work and dedication, including those who helped us test the engine before the release and submit bug reports.
```

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,178 @@
...
+
+![A list of projects with custom tags assigned to them](/storage/blog/godot-4-1-is-here/editor-project-manager-tags.png)
+
+### Rendering
+
+Particle turbulence got reworked in this version based on artists' feedback to offer greater creative control. Turbulence is used extensively to create these rich sprawling effects seen in many modern games. Big thanks to [KdotJPG](https://github.com/KdotJPG) and [Raffaele Picca](https://github.com/RPicster) for this contribution.
+
+![An effect driven by the updated particle turbulence system](/storage/blog/godot-4-1-is-here/render-particles-turbulence.png)
+
+Using the new 3D noise textures you can control the density of volumetric fog easily, and make it thinner in certain areas.
```
clayjohn:
```suggestion
Using the new NoiseTexture3D you can control the density of volumetric fog easily, and make it thinner in certain areas.
```

collections/_article/godot-4-1-is-here.md
YuriSizov: We are fully capable of making a guide similar to what we have for the 3 to 4 migration. It's not a lot of work, and Raul covered a big chunk of it.

collections/_article/godot-4-1-is-here.md
YuriSizov: What's wrong with using plain language here?

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,178 @@
...
+![An array of nodes exported as a custom property](/storage/blog/godot-4-1-is-here/editor-exported-node-array.png)
+
+This work was done by Tomasz and [Timothe Bonhoure](https://github.com/ajreckof)
+
+The project manager now allows you to assign tags to individual projects and filter projects by tags. It makes it much easier to search through dozens, if not hundreds of Godot projects.
+
+![A list of projects with custom tags assigned to them](/storage/blog/godot-4-1-is-here/editor-project-manager-tags.png)
+
+### Rendering
+
```
clayjohn, line 143:
```suggestion
### Rendering

While there are many [planned improvements](https://godotengine.org/article/rendering-priorities-4-1/) to Godot's renderers, most will be present in later releases. The rendering team (and all rendering contributors) prioritized bug fixing and stability over new features. Notably, the 3D GLES3 renderer is not yet complete, but will see substantial work over the coming months. 

```

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,186 @@
...
+More work will follow on the performance front throughout the year.
+
+### Core
+
+When importing models into Godot, there was often the problem that they ended up facing backward.
+
+![A model facing backwards (before the change)](/storage/blog/godot-4-1-is-here/model-facing-back.webp)
+
+[Juan Linietsky](https://github.com/reduz), [Tokage](https://github.com/TokageItLab), and [Aaron Franke](https://github.com/aaronfranke) addressed the issue by swapping the front and back camera directions in the editor. Also, the look_at() function now has an argument to use the model space as the reference for looking forward instead of the camera's minus z-axis. These changes also help fix a long-standing bug with path following.
```
akien-mga:
```suggestion
[Juan Linietsky](https://github.com/reduz), [Tokage](https://github.com/TokageItLab), and [Aaron Franke](https://github.com/aaronfranke) addressed the issue by swapping the front and back camera directions in the editor. Also, the `look_at()` function now has an argument to use the model space as the reference for looking forward instead of the camera's minus Z axis. These changes also help fix a long-standing bug with path following.
```

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,186 @@
...
+
+This version also introduces experimental multithreading for your scenes, a feature kickstarted by [Juan Linietsky](https://github.com/reduz). New node properties give you full control over how nodes get processed, sequentially or in parallel.
+
+![Thread group configuration options for Node](/storage/blog/godot-4-1-is-here/threaded-node-properties-highlighted.png)
+
+As mentioned, the feature is currently marked experimental as it needs extensive testing to find edge cases. We do not recommend using it in production yet. But it sets the foundation for making the most of modern hardware, with simple controls.
+
+Also on the multithreading front, [Pedro J. Estébanez](https://github.com/RandomShaper) worked hard on fixing a large amount of multi-threaded loading issues and limitations.
+
+On the rendering side, the Vulkan renderer got a shader cache to help prevent stalls due to shaders compiling at runtime. After the first compilation, the shader will be pulled directly from the cache. This cache can be toggled off in the project settings and is on by default in Godot 4.1.
```
clayjohn:
```suggestion
On the rendering side, the Vulkan renderer got a pipeline cache. While Godot already cached shaders to reduce shader compilation stutter, compiling pipelines still lead to some stuttering and slower load times. While the pipeline compilation stuttering issue is far from solved, this represents a step in the right direction and should also lead to a slight decrease in load times when using one of the RenderingDevice-based rendering backends. This cache can be toggled off in the project settings and is on by default in Godot 4.1.
```

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### GDExtension
+
+Godot comes with a unique technology to use low-level languages like C++ as a game scripting language, without having to recompile the engine.
+
+While the technology is still in beta for this release, GDExtension is now even closer to GDScript and C# in terms of scripting capabilities. You can now implement new visual shader nodes and create editor plugins with GDExtension.
+
+![An example of using a custom visual shader node from GDExtension](/storage/blog/godot-4-1-is-here/gdextension-visual-shader-node.png)
+
+The team also implemented a backward compatibility system to help ensure that code written for Godot 4.1 keeps working even if the API changes in future releases. Finally, a lot of work was done on the architecture once again to make the GDExtension API extensible in the future.
```
akien-mga:
```suggestion
The team also implemented a backward compatibility system to help ensure that code written for Godot 4.1 keeps working even if the API changes in future releases. However, a significant compatibility breakage was necessary to do in 4.1 to fix critical issues in GDExtension, so existing Godot 4.0 extensions will need to be ported and recompiled for 4.1.

Finally, a lot of work was done on the architecture once again to make the GDExtension API extensible in the future.
```

collections/_article/godot-4-1-is-here.md
```diff
@@ -0,0 +1,186 @@
...
+
+In addition to that, the editor will now keep track of your window layout so that when you close and reopen the editor, you should often find yourself exactly where you left off.
+
+You can thank [trollodel](https://github.com/trollodel), [Hendrik Brucker](https://github.com/Geometror), and [Tomasz Chabora](https://github.com/KoBeWi) for this.
+
+Godot 4.0 introduced the option to define and export typed arrays, and to export individual nodes to the inspector, but it was not possible to combine the two. From Godot 4.1 onwards, you can export arrays of nodes to the inspector, which is great to link game objects together.
+
+![An array of nodes exported as a custom property](/storage/blog/godot-4-1-is-here/editor-exported-node-array.png)
+
+This work was done by Tomasz and [Timothe Bonhoure](https://github.com/ajreckof)
```
akien-mga:
```suggestion
This work was done by Tomasz and [Timothé Bonhoure](https://github.com/ajreckof)
```

akien-mga (approved):
Looks good to me.

YuriSizov:
Thanks, everyone!

PR #667, opened 2023-07-10
Add new company - godot-consoles-all-you-need-know.md

honzapatCZ:
Hello
We have lately decided to invest in creating own ports of Godot for consoles.
As this article seems to be the most common place to find Godot porting info I thought it would be good idea to add ourselves here.

Not sure if there is maybe some footer EDIT note required?


YuriSizov:
Hey there! This is a blog post, it's not supposed to be edited like that. The correct place to add a porting company would be in the documentation: https://docs.godotengine.org/en/stable/tutorials/platform/consoles.html

That said, we would probably want to hear about Godot games that you have already published first. I hope you understand that this is important for the sake of the community that is going to read this article and see the link.

PR #677, opened 2023-07-18
Add XR Tools subteam and add Malcolm Nixon as member

BastiaanOlij:
XR Tools is our 1st party tool plugin that I created and have been maintaining for a few years now. @malcolmnixon for the last two years has been one of the main contributors and probably added more code to it then I have. Time to recognise that contribution.. :)


coppolaemilio:
Welcome!

PR #702, opened 2023-09-10
Add community forums

coppolaemilio:
This PR adds the community forums to the /community page.

It looks like this:

image


paddy-exe:
Shouldn't they first add a section about abiding the official Godot Code of Conduct before we link them on the website? I don't see anything related here (even though they do have some rules there which are similar/identical to the official ones): https://godotforums.org/p/2-rules

coppolaemilio:
Would be nice if they add a mention to our CoC, but the rules do seem to align pretty much.

YuriSizov (approved):

PR #709, opened 2023-09-19
Added note about the unexpected size of the project

babakfp:


Calinou:
Why does this have a significant importance by 2023 standards? A Git clone of godotengine/godot is already larger than that, especially after building.

babakfp:
Hi there
Sorry about the confusion.
Something weird is going on, the size of the project was 1.5GB, and now it's 2.7 without me doing anything. Let me check things out and I'll get back to you soon.

babakfp:
> Why does this have a significant importance by 2023 standards? A Git clone of godotengine/godot is already larger than that, especially after building.

Thank you for raising this point. You're absolutely right that, in the context of larger projects like godotengine/godot, a 2.71 gigabyte project size might not seem particularly significant. However, our intention in highlighting the project's size is to ensure that all contributors are aware of potential bandwidth and storage considerations, especially for those with limited resources or slower internet connections. While it may not be a major concern for everyone, it can still be beneficial information for some members of our community. We appreciate your input and aim to provide transparency to accommodate a diverse range of contributors and users.

YuriSizov:
> However, our intention in highlighting the project's size is to ensure that all contributors are aware of potential bandwidth and storage considerations, especially for those with limited resources or slower internet connections.

This can be done without colorful language, for starters.

Additionally, such a note must serve a practical purpose. Just telling people about the size doesn't do that. The codebase itself is only around 750 MB. But the size of a complete repo clone would inevitably be bigger simply because of all the history. If the cloned size is a concern, we can add a note with an advice on how to do a shallow clone that doesn't bring with itself a complete history from several years ago.

That would be a good and positive addition. But the note that you've added, unfortunately, is not.

babakfp:
> This can be done without colorful language, for starters.
> 
> Additionally, such a note must serve a practical purpose. Just telling people about the size doesn't do that. The codebase itself is only around 750 MB. But the size of a complete repo clone would inevitably be bigger simply because of all the history. If the cloned size is a concern, we can add a note with an advice on how to do a shallow clone that doesn't bring with itself a complete history from several years ago.
> 
> That would be a good and positive addition. But the note that you've added, unfortunately, is not.

Thank you for your feedback. I understand your perspective, and I appreciate your suggestions.

The intention behind the note about the project's size was indeed to raise awareness about potential bandwidth and storage considerations, especially for contributors who might have limited resources or slower internet connections. However, I apologize if the language used in the note seemed colorful or distracting to you; that was not the intention. English is not my native language.

You make a valid point about the size of the clone being massive due to its history. I'll work on revising the note to include more practical advice, such as how to do a shallow clone to address size concerns while also ensuring that it's communicated more constructively and positively.

PR #730, opened 2023-10-12
Added local user group of Karlsruhe (Germany) to the list of communities

nerdlibfront:
We have a started a Godot User Group in Karlsruhe (Germany) 2 month ago and it is flourishing since then, gaining more members every week.
It would be nice if the group can be listed on the Godot Webpage so that others from the region can join us.


coppolaemilio (approved):

coppolaemilio:
Welcome to the community!

PR #748, opened 2023-11-07
Add support for download config overrides

YuriSizov:
- Add Linux ARM builds for 4.2 beta5 and later
- Add an override for previous releases to remove Linux ARM builds.
- Add an override for early 4.0 alphas to remove non-existent .NET builds.

Overrides operate on inclusive version ranges. You can see how it looks in the modified YAML file. The first match config is going to be used, so order matters. This can be important if we introduce and remove something in a number of releases inside another range.

When testing I've checked that the download block in articles is unaffected, and so are download pages. Expected changes are:
- `/download/archive/4.2-beta5/` shows Linux ARM builds.
- `/download/archive/4.1.3-stable/` doesn't show Linux ARM builds.
- `/download/archive/4.0-alpha17/` shows the same as 4.1.3 stable and all other intermediate releases.
- `/download/archive/4.0-alpha16/` doesn't show C#/.NET builds.

To match versions to ranges I use a simple hashing mechanism which turns version name (e.g. "4.1.3") and version flavor (e.g. "alpha17") into integer values. These are easy to compare.

### DIFF
--- a/_data/download_configs.yml
+++ b/_data/download_configs.yml
@@ -5,6 +5,8 @@ defaults:
       android.apk: android_editor.apk
       linux.64: linux.x86_64.zip
       linux.32: linux.x86_32.zip
+      linux.arm64: linux.arm64.zip
+      linux.arm32: linux.arm32.zip
       macos.universal: macos.universal.zip
       windows.64: win64.exe.zip
       windows.32: win32.exe.zip
@@ -17,6 +19,8 @@ defaults:
       editor:
         linux.64: mono_linux_x86_64.zip
         linux.32: mono_linux_x86_32.zip
+        linux.arm64: mono_linux_arm64.zip
+        linux.arm32: mono_linux_arm32.zip
         macos.universal: mono_macos.universal.zip
         windows.64: mono_win64.zip
         windows.32: mono_win32.zip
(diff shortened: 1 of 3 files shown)


Calinou:
Tested locally, it works as expected:

image

Note that you need to click **Show all downloads** on the right for the ARM downloads to be visible (same goes for 32-bit x86, so this isn't a new issue).

PS: Is the ".NET - Standard" listing intended? I thought it was either "Standard" or ".NET", not both.

YuriSizov:
> Is the ".NET - Standard" listing intended? I thought it was either "Standard" or ".NET", not both.

It's an unfortunate necessity to make everything formatted properly across the website without having excessive configs.

It's just a .NET version of your normal editor build.

akien-mga:
> > Is the ".NET - Standard" listing intended? I thought it was either "Standard" or ".NET", not both.
> 
> It's an unfortunate necessity to make everything formatted properly across the website without having excessive configs.
> 
> It's just a .NET version of your normal editor build.

This could maybe be done like "Export templates" and "Export templates - .NET"?
We could also add "editor" like we do for "Web editor" so that it's clearer that users don't need to download say "Android" to get Android support.

So it would be:
```
Android editor - Universal APK (...)
Linux editor (x86_64)
Linux editor - .NET (x86_64)
Web editor
```

YuriSizov:
To be clear, this .NET Standard thing is already present on the website since many months ago. It was needed to unify some variations in the naming that we have. I didn't find a better solution and considered it a fine compromise.

akien-mga (approved):

YuriSizov:
Thanks for reviews, let's see if anything breaks 🙃

PR #766, opened 2023-12-12
Tweak phrase in 4.2.1 blog

JohnVeness:
Sorry, I missed the draft window. This is an attempt to correct "we still need confirmation that it's the case", which isn't good English. I went through "we still need confirmation that that's the case" (which is correct English but still a bit clunky) and a few others before settling on the simple version above. Could alternatively be "we still need confirmation about that". Thanks!


akien-mga (approved):

akien-mga:
Thanks!

PR #768, opened 2023-12-14
Rename "LTS" to "3.x LTS"

adamscott:
This PR renames "LTS" to "3.x LTS" as there's no such concept of LTS right now for Godot. 4.2 is the most stable 4.x release up-to-date and we should encourage people to use the 4.x branch instead of the aging 3.x one.

Hopefully, with this PR, we still say that our 3.x is well supported, but that the LTS term doens't implicate that 4.x isn't like long-term supported.

image

### DIFF
--- a/_layouts/download.html
+++ b/_layouts/download.html
@@ -49,7 +49,7 @@ layout: default
 			{% endfor %}
 
 			<p class="previous-releases">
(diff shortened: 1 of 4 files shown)


adamscott:
Superseded by #769.

PR #769, opened 2023-12-14
Rename LTS to 3.x

adamscott:
**_Supersedes #768_**

This PR renames "LTS" to "3.x" as there's no such concept of LTS right now for Godot. 4.2 is the most stable 4.x release up-to-date and we should encourage people to use the 4.x branch instead of the aging 3.x one.

We named 3.5 LTS as a way to signal that we still intended to support Godot 3.x, but LTS has a different meaning. People expect, with LTS, that the 3.5 branch will still be supported with security patches and such. But with the upcoming release of 3.6, it doesn't make sense to say that the LTS version is binded to either "3.x" or "3.6".

Hopefully, with this PR, we still say that our 3.x is well supported, but that the LTS term doesn't implicate that 4.x isn't long-term supported.

image

### DIFF
--- a/_layouts/download.html
+++ b/_layouts/download.html
@@ -49,7 +49,7 @@ layout: default
 			{% endfor %}
 
 			<p class="previous-releases">
-				<strong class="previous-releases-featured">For the LTS version, <a href="/download/3.x/windows" class="set-os-download-url" data-version="3">Download Godot 3</a>.</strong>
+				<strong class="previous-releases-featured">For the 3.x version, <a href="/download/3.x/windows" class="set-os-download-url" data-version="3">Download Godot 3</a>.</strong>
 				<br>
 				You can find previous releases in the <a href="/download/archive">download archive</a>.
 			</p>
--- a/pages/download/preview.html
+++ b/pages/download/preview.html
@@ -138,10 +138,10 @@ layout: default
 			<div class="download-hint">{{ stable_version_4.name }}</div>
 		</a>
 
-		<p>Want to stick to the <strong>true and trusted Godot 3</strong>? Download the long-term support version!</p>
+		<p>Want to stick to the <strong>true and trusted Godot 3</strong>? Download the latest 3.x version!</p>
 		<a href="/download/windows" class="btn btn-download set-os-download-url" data-version="3"
-			title="Download the long-term support version of Godot 3">
-			<div class="download-title">Download LTS</div>
+			title="Download the latest version of Godot 3">
+			<div class="download-title">Download 3.x</div>
 			<div class="download-hint">{{ stable_version_3.name }}</div>
 		</a>
 	</div>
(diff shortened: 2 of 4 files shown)


pages/download/preview.html
```diff
@@ -141,7 +141,7 @@ <h2>Need a stable version?</h2>
 		<p>Want to stick to the <strong>true and trusted Godot 3</strong>? Download the long-term support version!</p>
```
JohnVeness: Should this line change too?

adamscott: Yes! Forgot to check about the text format of LTS.

coppolaemilio (approved):

JohnVeness:
Not in scope for this PR, as the source is in a different repo, but I notice "LTS" is still mentioned in docs.godotengine.org, e.g. https://docs.godotengine.org/en/latest/about/release_policy.html in reference to 3.6.

_layouts/download.html
```diff
@@ -49,7 +49,7 @@ <h1>Download Godot&nbsp;4 for {{ page.platform }}</h1>
 			{% endfor %}
 
 			<p class="previous-releases">
-				<strong class="previous-releases-featured">For the LTS version, <a href="/download/3.x/windows" class="set-os-download-url" data-version="3">Download Godot 3</a>.</strong>
+				<strong class="previous-releases-featured">For the 3.x version, <a href="/download/3.x/windows" class="set-os-download-url" data-version="3">Download Godot 3</a>.</strong>
```
YuriSizov, line 52: Isn't this a weird phrase now? :)

pages/download/preview.html
```diff
@@ -138,10 +138,10 @@ <h2>Need a stable version?</h2>
 			<div class="download-hint">{{ stable_version_4.name }}</div>
 		</a>
 
-		<p>Want to stick to the <strong>true and trusted Godot 3</strong>? Download the long-term support version!</p>
+		<p>Want to stick to the <strong>true and trusted Godot 3</strong>? Download the latest 3.x version!</p>
```
YuriSizov, line 141: And this as well (and same in Features)

YuriSizov:
As mentioned in chat, this is not an urgent thing to change. I would rather we addressed the bigger problem with people not knowing why they would pick 4 over 3, or vice versa. I think we would make things slightly worse if we remove the LTS label to replace it with "3.x" because at least calling it LTS set up some expectations.

I never liked the use of LTS, just couldn't find anything better which would be as succinct. But at this point the solution might just be to de-prioritize 3.x downloads throughout. And we definitely need a page comparing both versions which we can link where relevant. But I wouldn't do just the rename, it's not immediately necessary.

winston-yallow:
Just as an addition consideration:

At least on discord we often have questions like "is Godot 3 still supported?" to which people have often answered "yes, because it is LTS". So removing the LTS could signal "now it's not supported anymore".

Having a page up explaining the versions would be great before changing the name.

coppolaemilio:
Closed by https://github.com/godotengine/godot-website/commit/bb0f0a41c3cb1a69dbf7f96ce9c4d98eb9e55822 and https://github.com/godotengine/godot-website/pull/840.

Thanks!

PR #770, opened 2023-12-16
Link to up-to-date information in per-instance uniforms article

Calinou:
This article regularly comes up in search results, so it should point to up-to-date documentation.

See https://forum.godotengine.org/t/adjust-shaders-per-instance/37948.

PS: We should probably add a reusable block of markup we can use for those "Update" blocks in blog posts, so they can look more uniform across articles and could be styled differently.


coppolaemilio (approved):

coppolaemilio:
Thanks!

PR #781, opened 2024-01-11
Fix header links padding

MichaelJRM:
Fixes the small padding imperfections in the header links when hovering

Before
before

After
after

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,6 +1,6 @@
 <input type="checkbox" id="nav_toggle_cb">
 <header>
-	<div class="container flex align-center">
+	<div class="container flex">
 		<div id="nav_head">
 			<a href="/" id="logo-link">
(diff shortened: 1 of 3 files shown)


YuriSizov:
Hey, thanks for your contribution! Please omit changing Jekyll's version from this PR, it should be done separately.

MichaelJRM:
> Hey, thanks for your contribution! Please omit changing Jekyll's version from this PR, it should be done separately.

Hey, thanks for the feedback! This is my first time using Jekyll and I was getting the error bellow, changing the version fixed it but I ended up forgetting to revert it :D

```
Ruby33-x64/lib/ruby/3.3.0/logger.rb:384:in `level': undefined method `[]' for nil (NoMethodError)

@level_override[Fiber.current] || @level
```

PR #784, opened 2024-01-18
Update showcase page for Until Then

semickolon:
Please let me know if there's anything you want changed or corrected. Thank you!


coppolaemilio:
Thanks! Changes look good, made some edits to also feature you on the home page. 
The release date has to be "TBD" for it to show properly, but I will see to update how that template works to make it so we can display a tentative date in another commit.

PR #786, opened 2024-01-18
Fix missing communities on the map in user groups page

### LINKED ISSUE #785: User groups missing on the map
Some user groups are missing in the user group map, ie. all countries from Iraq to the end of the list (the United States).

image

The following error in the web console seems related.
```
Uncaught TypeError: t is undefined
    project Projection.SphericalMercator.js:24
    latLngToPoint CRS.js:28
    project Map.js:993
    latLngToLayerPoint Map.js:1015
    update Marker.js:194
    onAdd Marker.js:113
    _layerAdd Layer.js:114
    whenReady Map.js:1465
    addLayer Layer.js:176
    addTo Layer.js:52
    <anonymous> (index):217
    <anonymous> (index):202
Projection.SphericalMercator.js:24:35
``

### PR DESCRIPTION
**Pull request by adamscott**
This PR adds Iraq to the list of countries coordinates. The fact that it was missing was crashing the mapping process.

Fixes #785

### DIFF
--- a/_data/communities.yml
+++ b/_data/communities.yml
@@ -221,7 +221,6 @@ Iraq:
         url: https://www.instagram.com/bgl.iq/
       - title: Linkedin
         url: https://www.linkedin.com/company/bgh-iq
-
 Israel:
   - name: Godot Israel
     links:
--- a/pages/community/user-groups.html
+++ b/pages/community/user-groups.html
@@ -44,6 +44,7 @@ layout: default
 				'Indonesia': [-2.548926, 118.0148634],
 				'Italy': [42.504154, 12.646361],
 				'Iran': [32.4207423, 53.6830157],
+				'Iraq': [33.223191, 43.679291],
 				'Israel': [31.53131, 34.86677],
 				'Japan': [36.57484, 139.23942],
 				'Latvia': [56.959065, 24.862593],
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: _data/communities.yml [hunk 1]
```diff
@@ -221,7 +221,6 @@ Iraq:
         url: https://www.instagram.com/bgl.iq/
       - title: Linkedin
         url: https://www.linkedin.com/company/bgh-iq
-
```
adamscott, line 224: Unrelated, purely cosmetic, but I included it in the PR.

pages/community/user-groups.html
```diff
@@ -44,6 +44,7 @@ <h1 class="intro-title">User Groups</h1>
 				'Indonesia': [-2.548926, 118.0148634],
 				'Italy': [42.504154, 12.646361],
 				'Iran': [32.4207423, 53.6830157],
+				'Iraq': [33.223191, 43.679291],
```
YuriSizov, line 47: I think we should add a check and handle missing entries in the list more gracefully. We should still log an error, but we shouldn't break and stop rendering the remainder of the list if some country is missing.

adamscott: Added a check:
image

pages/community/user-groups.html
```diff
@@ -139,6 +140,11 @@ <h1 class="intro-title">User Groups</h1>
 
 				// Add markers for online communities with merged tooltips.
 				Object.keys(onlinePoints).forEach((country, index) => {
+					if (countryCoordinates[country] == null) {
+						console.warn(`Coordinates of \"${country}\" are missing, skipping.`);
+						return;
+					}
+
```
adamscott: @YuriSizov Here's the check that I just added, following your comment!

YuriSizov (approved):

pages/community/user-groups.html
adamscott: I just pushed a new version where the backslashes were removed, as they are not needed anymore.

coppolaemilio:
Merci!

PR #796, opened 2024-02-09
Convert showcase images to webp and improve image rendering

MichaelJRM:
Sorry for the large pr, it's mostly changes from .png, .jpg to webp

- Fixed an issue with background bleeding when using rounded borders
- Converted all showcase images to webp
- Made all showcase images lazy load
- Fixed images size to prevent layout shifting

**Background bleeding before**
before
**After**
after


_includes/footer.html
```diff
@@ -47,24 +47,24 @@
...
-					<img src="/assets/footer/reddit_logo.svg" width="32" height="32" alt="Red​dit">
+					<img src="/assets/footer/reddit_logo.svg" width="32" height="32" alt="Reddit" loading="lazy">
 				</a>
 				<a href="/rss.xml" target="_blank" rel="noopener">
 					<!-- Icon is called `feed` instead of `rss` to prevent content blockers from blocking the icon -->
-					<img src="/assets/footer/feed_logo.svg" width="32" height="32" alt="RSS feed">
+					<img src="/assets/footer/feed_logo.svg" width="32" height="32" alt="RSS feed" loading="lazy">
 				</a>
 			</div>
 		</div>
```
Calinou: I wouldn't use lazy-loading for SVGs as these files are tiny, yet seeing them with a visible delay feels odd.

MichaelJRM: Yes makes sense, I've removed it & also I didn't notice the comment about the zero-width space, fixed as well.

_layouts/showcase-item.html
```diff
@@ -183,16 +195,15 @@ <h3 style="opacity: 0.75; margin-top: 0.5rem">
...
 				</iframe>
 			</div>
 			{% endif %}
 
 			{% for image in page.gallery %}
 			<a href="{{ image }}" class="lightbox" data-group="showcase">
 				<img class="rounded-lg" src="{{ image }}" alt="Screenshot of {{ page.title }} by {{ page.author }}"
-					style="width: 100%; height: auto; margin-top: 1rem; background:  linear-gradient(90deg, #947451 12%, #a8885f 29%, #8c8977 65%, #d2ba9f 98%) "
-					width="16" height="9">
+					width="16" height="9" onload="removeFeatureImgBackground(this)">
```
coppolaemilio: Instead of doing this with JavaScript I would simply add 
`style="background: unset; background-color: unset; border: unset;"` to that element instead of the `onload`

coppolaemilio: Did a test myself and with just `style="background: transparent;"` it does the trick

coppolaemilio: If you do that also with the home page change we should be good to merge! thanks (:

MichaelJRM: Isn't the background used as a placeholder to show there is an image loading?
image-loading

If we remove the background then there is no indication that there is supposed to be an image here before the image actually appearing
image-loading-without-background

coppolaemilio: True! missed that since it was loading so fast locally... I'm not sure is super necessary to make the progressive loading happen, specially now that the images are webp and load faster 😅

Calinou: WebP can still load slowly if on a slow mobile network (3G/4G with bad coverage), so placeholders are still important.

WebP also doesn't support progressive loading like JPEG (AVIF doesn't support that either), so placeholders aren't going away anytime soon… Only JPEG-XL supports progressive loading like JPEG.

MichaelJRM: I think it makes sense to leave the placeholder, even tho is faster now it still improves the experience for some users.

I kinda sidetracked a bit and ended up optimizing other stuff unrelated to images, I promise I won't change anything else. Loading times are much faster now and touching 100 performance score :D

I found a gem called jekyll_picture_tag that helped a lot, I've used it where it made sense.

coppolaemilio:
Hey, thanks for all the efforts optimizing the website! Unfortunately, I feel like the PR is a bit out of scope now.
I would recommend to make the changes in different PRs. The Images conversions to webp would be a super easy merge, but the rest of the changes require individual discussions. If you want to coordinate better with the rest of the team, please make sure to join the dev chat at: https://chat.godotengine.org/channel/website

goodusername123:
@MichaelJRM
Hey if possible please set [`smart_subsample`](https://www.ctrl.blog/entry/webp-sharp-yuv.html) to `true` and `effort` to `6` for all of the WebPs setup in `picture.yaml` ([relevant doc](https://rbuchberger.github.io/jekyll_picture_tag/users/presets/image_quality.html#image-format-options))

The blog I linked for "smart_subsample" explains why it's important, And for `effort` the default value is `4` while higher values like `5` and `6` enable [Trellis quantization](https://en.wikipedia.org/wiki/Trellis_quantization) which is something that's enabled by default in encoders for other formats like [MozJPEG](https://github.com/mozilla/mozjpeg) and benefits file-size and quality.

MichaelJRM:
@coppolaemilio I've created a new pr https://github.com/godotengine/godot-website/pull/798 with only the image conversion stuff, I'll create new ones for the other features and close this one.

Thanks for the tip  @goodusername123 I didn't notice it when reading the docs.

PR #801, opened 2024-02-13
Directed acyclic graph for rendering device article.

DarioSamo:
I've finished my first writing pass on the article. Feel free to point out any weirdness in the writing as I'm not a native English speaker so it's a bit harder to tell if something sounds odd.

It's also very dense so I could've forgotten to introduce a concept that people might not be familiar with, although the expected knowledge of the reader should be someone who's at least aware of how graphics in games work (hence why a short paragraph gives a summary of the feature and later does a deep dive into the implementation for the readers who are interested).

TODO:

- [x] Cover image (I have no idea what to use as one for this at the moment).


collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+No changes are expected from users whatsoever. If you're one of the few people who have written code that uses Rendering Device, there's no need to worry either: while the API has changed slightly, all of the methods just require less information than before. What level of performance improvement will very much depend on the contents of the Godot project. For example, GPU particle systems will get massive improvements, while more standard scenes that use post-processing will see some frametime reduction in the ballpark of around 5% to 15%. On top of that, the Godot developers will have a much easier time improving the performance of the renderers in the near future.
+
+If you're interested in the details of how this was achieved, keep on reading.
+
+*Disclaimer: Darío, the author of this article, is a new face around this blog who was hired by **W4 Games** last summer to start contributing to the project. His work is completely sponsored by **W4 Games** and you can find more of his contributions to the engine in his [GitHub profile](https://github.com/DarioSamo).*
```
JohnVeness: I suggest moving this paragraph to the top of the article, and removing the word "Disclaimer".

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
+---
+title: "GPU synchronization in Godot 4.3 is getting a major upgrade"
+excerpt: "Say goodbye to placing barriers, hello to the new acyclic graph"
+categories: ["progress-report"]
+author: Darío Banini
+image: /storage/blog/covers/todo.webp
+date: 2024-02-14 9:00:00
+---
+
+Since the introduction of Godot 4, [Rendering Device](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) has been the backbone of the Forward+ and Mobile renderers. Making APIs that are both easy to use and flexible enough to cover all the features users want in a game engine is a very difficult task. As far as GPU synchronization is concerned, the high level of detail expected by APIs like Vulkan or D3D12 compared to its predecessors is worthy of a few blog posts of its own. This article will try to keep the topic of synchronization as brief as possible to focus on the problems solved within Godot instead.
```
JohnVeness:
```suggestion
Since the introduction of Godot 4, [Rendering Device](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) has been the backbone of the Forward+ and Mobile renderers. Making APIs that are both easy to use and flexible enough to cover all the features users want in a game engine is a very difficult task. As far as GPU synchronization is concerned, the high level of detail expected by APIs like Vulkan or Direct3D 12 compared to their predecessors is worthy of a few blog posts of its own. This article will try to keep the topic of synchronization as brief as possible to focus on the problems solved within Godot instead.
```

I know it's linked, so someone can press it for more info, but I think it would be nice to say "the Rendering Device API" or "the Rendering Device class" or similar in its first mention. Also, should it be RenderingDevice with no space?

I also found it slightly jarring that GPU synchronization is mentioned in the middle of the paragraph without any explanation as to what that is.

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+With the goal of reducing the difficulty of the development of the new renderers and allowing the GPU to parallelize work more effectively, the automatic construction of a [directed acyclic graph (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph) during rendering has been introduced at the lowest levels of the engine. [This change is already merged](https://github.com/godotengine/godot/pull/84976) in the `main` development branch of Godot and should be part of the **4.3 release**. The introduction of the graph will introduce both performance improvements and various bug fixes for issues that were very difficult to investigate. For example, after the graph was merged it was discovered that MSAA with SSAO will no longer cause artifacts in AMD Polaris ([#61415](https://github.com/godotengine/godot/issues/61415)), despite no effort being spent towards developing a specific fix for the problem.
```
JohnVeness:
```suggestion
With the goal of reducing the difficulty of the development of the new renderers and allowing the GPU to parallelize work more effectively, the automatic construction of a [directed acyclic graph (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph) during rendering has been introduced at the lowest levels of the engine. [This change is already merged](https://github.com/godotengine/godot/pull/84976) in the `main` development branch of Godot and should be part of the **4.3 release**. The introduction of the graph will bring both performance improvements and various bug fixes for issues that were very difficult to investigate. For example, after the graph was merged it was discovered that MSAA with SSAO will no longer cause artifacts in AMD Polaris ([#61415](https://github.com/godotengine/godot/issues/61415)), despite no effort being spent towards developing a specific fix for the problem.
```
To avoid "introduction" and "introduce" in close proximity.

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+No changes are expected from users whatsoever. If you're one of the few people who have written code that uses Rendering Device, there's no need to worry either: while the API has changed slightly, all of the methods just require less information than before. What level of performance improvement will very much depend on the contents of the Godot project. For example, GPU particle systems will get massive improvements, while more standard scenes that use post-processing will see some frametime reduction in the ballpark of around 5% to 15%. On top of that, the Godot developers will have a much easier time improving the performance of the renderers in the near future.
```
JohnVeness:
```suggestion
No changes are expected from users whatsoever. If you're one of the few people who have written code that uses Rendering Device, there's no need to worry either: while the API has changed slightly, all of the methods just require less information than before. What level of performance improvement you see will very much depend on the contents of your project. For example, GPU particle systems will get massive improvements, while more standard scenes that use post-processing will see some frametime reduction in the ballpark of around 5% to 15%. On top of that, the Godot developers will have a much easier time improving the performance of the renderers in the near future.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Background
+
+Understanding the problem space is crucial to do a deep dive into the technical decisions that were made. Dealing with an existing codebase for a general engine that had thousands of lines written on top of an existing API imposes many restrictions on what sort of changes are allowed. Flaws introduced in the early stages of design can have long term effects on the development of a big project. This was very much the case with some of the key decisions taken for the Rendering Device abstraction, its coupling to Vulkan and the negative results on the rest of the codebase. The good news is it's never too late to go back to the drawing board. [reduz](https://github.com/reduz) laid out the plans for this [redesign](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08) during the end of the previous year and the team got to work on how to make it a reality.
```
JohnVeness:
```suggestion
Understanding the problem space is crucial to do a deep dive into the technical decisions that were made. Dealing with an existing codebase for a general engine that had thousands of lines written on top of an existing API imposes many restrictions on what sort of changes are allowed. Flaws introduced in the early stages of design can have long-term effects on the development of a big project. This was very much the case with some of the key decisions taken for the Rendering Device abstraction, its coupling to Vulkan and the negative results on the rest of the codebase. The good news is it's never too late to go back to the drawing board. [reduz](https://github.com/reduz) laid out the plans for this [redesign](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08) during the end of 2023 and the team got to work on how to make it a reality.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Vulkan
+
+There's a few Vulkan concepts that must be understood before digging into why Rendering Device was designed the way it was for Godot 4 and the difficulties it encountered along the way.
+
+#### Command Buffers
```
JohnVeness:
```suggestion
#### Command buffers
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+Recording work for the GPU and submitting it for execution is a very explicit operation in Vulkan. [Command buffers](https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html) (or command lists in D3D12) are the objects where all the recorded work is stored that that will executed at a later point on the GPU. These buffers can grow as much as the user desires and they won't be run until they're submitted for execution to a command queue. This implies it's essentially possible to record from multiple sources, even multiple threads, and submit the work to the GPU once it's ready with proper synchronization.
+
+#### Render Passes
```
JohnVeness:
```suggestion
#### Render passes
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+Drawing something with Vulkan requires a lot of information upfront in the form of a "render pass": an object that contains references to the textures that the GPU will use as targets, their initial and final states (texture memory layouts) and much more. This is a stark contrast from older APIs like OpenGL or even D3D12, where the render target can be changed during command recording and does not require the creation of an object ahead of time. While a render pass is active, there's also various restrictions on what operations can be recorded on the command buffer. The render pass must end before being allowed to use all types of operations again.
```
JohnVeness:
```suggestion
Drawing something with Vulkan requires a lot of information upfront in the form of a "render pass": an object that contains references to the textures that the GPU will use as targets, their initial and final states (texture memory layouts) and much more. This is a stark contrast from older APIs like OpenGL or even D3D12, where the render target can be changed during command recording and does not require the creation of an object ahead of time. While a render pass is active, there are also various restrictions on what operations can be recorded on the command buffer. The render pass must end before being allowed to use all types of operations again.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### Barriers
+
+The biggest change compared to previous APIs is that synchronization between commands is no longer automatically deduced by the driver and it must be manually implemented by the programmers using Vulkan. If you want to read more about this topic in detail, I highly recommend Hans-Kristian's [blog post about Vulkan Synchronization](https://themaister.net/blog/2019/08/14/yet-another-blog-explaining-vulkan-synchronization/) and reading the [official Vulkan documentation](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdPipelineBarrier.html). To keep the article short, a very basic explanation is provided below.
```
JohnVeness:
```suggestion
The biggest change compared to previous APIs is that synchronization between commands is no longer automatically deduced by the driver and instead must be manually implemented by the programmers using Vulkan. If you want to read more about this topic in detail, I highly recommend Hans-Kristian's [blog post about Vulkan Synchronization](https://themaister.net/blog/2019/08/14/yet-another-blog-explaining-vulkan-synchronization/) and reading the [official Vulkan documentation](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdPipelineBarrier.html). To keep the article short, a very basic explanation is provided below.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+The order of execution of the recorded commands inside a command buffer **is NOT guaranteed to complete in the order they were submitted in**: the GPU can reorder these commands in whatever order it thinks is best to complete the job as quickly as possible. To compensate for this, the programmer must manually insert **[synchronization barriers](https://gpuopen.com/learn/vulkan-barriers-explained/)** in the command buffer that allow specifying in detail which commands should be completed or start by specifying the scope **before** and **after**. The scope includes multiple concepts such as the execution stage (e.g. drawing, compute, transfer), the type of access (e.g. read or write) and the affected memory regions (globally, buffers or textures). On top of that, barriers are capable of transitioning textures from one memory layout to another, which is a requirement to be able to use textures in different commands optimally.
```
JohnVeness:
```suggestion
The order of execution of the recorded commands inside a command buffer **is NOT guaranteed to complete in the order they were submitted**: the GPU can reorder these commands in whatever order it thinks is best to complete the job as quickly as possible. To compensate for this, the programmer must manually insert **[synchronization barriers](https://gpuopen.com/learn/vulkan-barriers-explained/)** in the command buffer that allow specifying in detail which commands should be completed or start by specifying the scope **before** and **after**. The scope includes multiple concepts such as the execution stage (e.g. drawing, compute, transfer), the type of access (e.g. read or write) and the affected memory regions (globally, buffers or textures). On top of that, barriers are capable of transitioning textures from one memory layout to another, which is a requirement to be able to use textures in different commands optimally.
```

JohnVeness: I'm not sure about "which commands should be completed or start" - should it be "which commands should be completed or started"?

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+| ![](/storage/blog/acyclic-graph/barriers.webp) |
+|:--:| 
+| *A barrier can establish a dependency between commands by specifying a synchronization scope.* |
+
+Beginners will have a hard time understanding this mechanism, but even experts or IHVs are not safe from making mistakes. A run of the [Vulkan Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers) in *synchronization* mode will reveal multiple issues on most Godot projects or even commercial games available in the market. These are among the most frustrating issues to understand as they won't even appear consistently depending on hardware vendor or hardware speed. Since eliminating these problems was one of the main goals behind the introduction of the acyclic graph, extensive use was made of these validation tools to ensure that no synchronization errors remained.
```
JohnVeness:
```suggestion
Beginners will have a hard time understanding this mechanism, but even experts or hardware vendors are not safe from making mistakes. A run of the [Vulkan Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers) in *synchronization* mode will reveal multiple issues on most Godot projects or even commercial games available in the market. These are among the most frustrating issues to understand as they won't even appear consistently depending on hardware vendor or hardware speed. Since eliminating these problems was one of the main goals behind the introduction of the acyclic graph, extensive use was made of these validation tools to ensure that no synchronization errors remained.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Rendering Device
+
+[This is the abstraction](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) on which the *Forward+* and *Mobile* renderers in Godot were built. It exposes an interface that provides a reasonable level of control over the GPU in a way that is not as fine-grained as Vulkan or D3D12 but is also not as "stateful" as OpenGL. More of the rendering state needs to be defined in advance, and the chances of causing an error by leaking state from one previous command to the next are significantly reduced.
```
JohnVeness:
```suggestion
[Rendering Device is the abstraction](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) on which the *Forward+* and *Mobile* renderers in Godot were built. It exposes an interface that provides a reasonable level of control over the GPU in a way that is not as fine-grained as Vulkan or D3D12 but is also not as "stateful" as OpenGL. More of the rendering state needs to be defined in advance, and the chances of causing an error by leaking state from one previous command to the next are significantly reduced.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+While Rendering Device attempted to hide many of the difficulties that Vulkan introduced, there are a few that slipped through that made writing rendering code for Godot 4 more difficult than anticipated.
+
+#### Draw list actions
+
+Just as draw lists were naturally mapped to render passes, so too was the fact that the start and end action had to be specified for the textures associated with them. These actions cover operations such as loading the contents of a texture, discarding it or clearing it with an specific color. There's plenty of performance to be gained by choosing the optimal option in each render pass. However, due to some internal design decisions, the introduction of a "CONTINUE" action was made that doesn't actually exist in Vulkan. This action essentially meant "the previous draw list is compatible and the texture should not be transitioned to a different state". For an engine like Godot, which must provide a lot of **optional** post-processing effects and drawing layers, this proved to be a *maintenance nightmare*, ending in a ton of branching in the code that was not very easy to follow and prone to mistakes.
```
JohnVeness:
```suggestion
Just as draw lists were naturally mapped to render passes, so too was the fact that the start and end action had to be specified for the textures associated with them. These actions cover operations such as loading the contents of a texture, discarding it or clearing it with a specific color. There's plenty of performance to be gained by choosing the optimal option in each render pass. However, due to some internal design decisions, the introduction of a "CONTINUE" action was made that doesn't actually exist in Vulkan. This action essentially meant "the previous draw list is compatible and the texture should not be transitioned to a different state". For an engine like Godot, which must provide a lot of **optional** post-processing effects and drawing layers, this proved to be a *maintenance nightmare*, ending in a ton of branching in the code that was not very easy to follow and prone to mistakes.
```

JohnVeness (commented):
Nice article, although I've only read a bit of it. Submitting a few suggestions so far :)

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+However, being able to reorder commands meant that an intermediate step had to be introduced where commands were recorded into an auxiliary structure that could be reordered and then converted to the corresponding native API commands. One possibility was encoding the Vulkan command arguments into the auxiliary buffer, but that approach meant the entire graph structure and logic would need to be implemented for every other backend as well. Therefore, it was deemed it'd be necessary for Pedro to work on his [pull request](https://github.com/godotengine/godot/pull/83452) that introduces an abstract interface for all the supported graphic APIs, including Vulkan, D3D12 and Metal in the near future. Thanks to this change, it was possible to use a single abstract API to use for encoding commands into the auxiliary buffer.
```
JohnVeness: The last sentence could do with a change so that there's not two instances of "to use".

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+The initial redesign was laid out in *reduz*'s [draft](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08), which was largely inspired by [Pavlo Muratov's "Organizing GPU Work with Directed Acyclic Graphs"](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) and showed the possibility of how the concept could be applied to Godot's existing design. Not everything stated in the document made it to the final version: in fact, the changes to Rendering Device were much less severe than initially indicated and the interface remained largely compatible. While the article that was used as inspiration includes a very detailed algorithm in how to implement multi-queue submission by using the graph, the team made the decision to cut this idea short and stick to a single command queue to begin with, as the difficulty of the task would come from building the graph automatically and would already take a significant amount of development time.
```
JohnVeness:
```suggestion
The initial redesign was laid out in *reduz*'s [draft](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08), which was largely inspired by [Pavlo Muratov's "Organizing GPU Work with Directed Acyclic Graphs"](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) and showed the possibility of how the concept could be applied to Godot's existing design. Not everything stated in the document made it into the final version: in fact, the changes to Rendering Device were much less severe than initially indicated and the interface remained largely compatible. While the article that was used as inspiration includes a very detailed algorithm in how to implement multi-queue submission by using the graph, the team made the decision to cut this idea short and stick to a single command queue to begin with, as the difficulty of the task would come from building the graph automatically and would already take a significant amount of development time.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Solution
```
JohnVeness: There are two level 2 headings called "Solution"!

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+- It is never possible for slices on the dirty list to overlap. In fact, these ranges are represented as 2D rectangles, which is pretty fun way to represent the problem visuallly!
+- The tracker for a texture slice holds a reference to its parent and a flag to identify whether it is present on the dirty list or not.
+- It is not possible for one command to use slices of the same texture that overlap, as this can lead to undefined behavior. This restriction is applied even if their usage is the same.
+- If the parent texture is used in a command directly, all slices in the dirty list will be reverted to the usage of the parent by "normalizing" the texture to one memory layout. It is not possible to use slices with a different usage during this step.
+- Slices present in the dirty list can also be normalized and removed individually if a new command wants to use a slice that'd overlap with their dirty region. Since it's not possible for slices to overlap in the same command, this operation can always be performed safely if the slice being normalized was used in an older command.
```
JohnVeness:
```suggestion
- Slices present in the dirty list can also be normalized and removed individually if a new command wants to use a slice that would overlap with their dirty region. Since it's not possible for slices to overlap in the same command, this operation can always be performed safely if the slice being normalized was used in an older command.
```
Although "that'd" is a real word, it's not often used in semi-formal documents like this!

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,279 @@
...
+- Dependency tracking is performed strictly with the tracker associated to the parent texture of a slice and not the slice's tracker, as it usually leads to the safest behavior.
```
JohnVeness:
```suggestion
- Dependency tracking is performed strictly with the tracker associated to the parent texture of a slice and not the slice's tracker, as this usually leads to the safest behavior.
```

collections/_article/rendering-acyclic-graph.md
DarioSamo: I switched around between both and Rendering Device with the space looked better to me.

I agree the mention of synchronization isn't even necessary on this paragraph. I think explaining the concept would detract from the purpose of giving a very short overview of what the article is about so readers know if they're interested in diving into the topic... so I think removing the mention actually makes the paragraph better and works for the same purpose.

collections/_article/rendering-acyclic-graph.md
DarioSamo: I actually wrote what you said at first and then making it go through google translate gave me this sentence instead. But what you said sounds more correct to me as well.

collections/_article/rendering-acyclic-graph.md
DarioSamo: Done.

DarioSamo:
Thank you very much @JohnVeness, applied all of your suggestions.

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+The other major benefit is that a lot of hard to identify bugs that were caused due to synchronization bugs are now potentially fixed. As pointed out before, the [MSAA with SSAO issue](https://github.com/godotengine/godot/issues/61415) was resolved as a side-effect. Problems like this one proved to be extremely hard to fix as Godot developers would require the particular set of hardware and the right scenarios for the defects to trigger. With the graph sufficiently complete and unable to cause these synchronization problems, the possibility of introducing errors of this kind disappears.
+
+### GPU performance
+
+The results are generally positive. No performance regressions have been identified in any scene as far as GPU performance is concerned, and in virtually most scenarios an improvement can be expected depending on their contents.
+
+| ![](/storage/blog/acyclic-graph/benchmark-projects.png) |
+|:--:| 
+| *Results gathered from running various projects with an NVIDIA GeForce RTX 3090 Ti at 1920x1080.* |
```
JohnVeness:
```suggestion
| *Results gathered from running various projects with an NVIDIA GeForce RTX 3090 Ti at 1920x1080. Higher is better.* |
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+* Legend of the Nuku Warriors: Internal demo scene by W4 Games.
+* [Desert Light Demo](https://github.com/RPicster/godot4-demo-desert-light) by [RPicster](https://github.com/RPicster).
+* [Reflection for Godot 4.0](https://github.com/Calinou/godot-reflection) ported by [Calinou](https://github.com/Calinou).
+* [Abandoned Spaceship Demo](https://github.com/perfoon/Abandoned-Spaceship-Godot-Demo) by [Perfoon](https://github.com/perfoon).
+
+The biggest gains by far could be identified in projects with multiple GPU particle systems, where the execution can now take place in parallel wherever it's possible. A dedicated particles benchmark makes this difference much more obvious.
+
+| ![](/storage/blog/acyclic-graph/benchmark-particles.png) |
+|:--:| 
+| *Results gathered from running the benchmark with NVIDIA GeForce RTX 3090 Ti.* |
```
JohnVeness:
```suggestion
| *Results gathered from running the benchmark with NVIDIA GeForce RTX 3090 Ti. Higher is better.* |
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+So why is this not enabled in `master` yet? It seems that secondary command buffers are prone into running into some strange issues on different hardware and are not as widely supported as they should be. During testing an issue was found in NVIDIA GPUs where the editor window would just go completely blank if secondary buffers were used under certain conditions. Apparently it wouldn't be the smartest decision to rely on this feature and expect it to work correctly on most platforms. However, if you want to experiment with it yourself [the code is currently present behind a compilation macro](https://github.com/godotengine/godot/blob/d3352813ea44447bfbf135efdec23acc4d1d3f89/servers/rendering/rendering_device.cpp#L59). The code will automatically enable multithreaded recording when the draw lists are determined to be large enough to be suitable and will result in a real reduction in CPU times in demanding scenes, provided the user has extra CPU cores to handle it.
```
JohnVeness:
```suggestion
So why is this not enabled in `master` yet? It seems that secondary command buffers are prone to running into some strange issues on different hardware and are not as widely supported as they should be. During testing an issue was found in NVIDIA GPUs where the editor window would just go completely blank if secondary buffers were used under certain conditions. Apparently it wouldn't be the smartest decision to rely on this feature and expect it to work correctly on most platforms. However, if you want to experiment with it yourself [the code is currently present behind a compilation macro](https://github.com/godotengine/godot/blob/d3352813ea44447bfbf135efdec23acc4d1d3f89/servers/rendering/rendering_device.cpp#L59). The code will automatically enable multithreaded recording when the draw lists are determined to be large enough to be suitable and will result in a real reduction in CPU times in demanding scenes, provided the user has extra CPU cores to handle it.
```
Actually, I think something simpler like "secondary command buffers can run into some" would be even better.

Also, right at the end, I think "enough CPU cores" sounds better, but I'm not sure if that has the correct meaning.

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+An alternative being evaluated is to record multiple primary command buffers instead and chain them together when the frame ends. However, submitting a command buffer for execution has a fixed cost that isn't insignificant according to IHVs, so some consideration must be given to creating different command buffers only when the benefit outweighs the cost.
```
JohnVeness:
```suggestion
An alternative being evaluated is to record multiple primary command buffers instead and chain them together when the frame ends. However, submitting a command buffer for execution has a fixed cost that isn't insignificant according to hardware vendors, so some consideration must be given to creating different command buffers only when the benefit outweighs the cost.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Future work
+
+With the introduction of the directed acyclic graph and with a [few more abstraction rewrites coming down the line](https://github.com/godotengine/godot/pull/87340), the engine will now have access to even more optimizations that can be implemented in the future.
+
+### Multiple queues
+
+Pavlo Muratov's [article](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) that was used as the main inspiration behind this change contains a very interesting proposal in how to submit and synchronize GPU work across multiple the queues exposed by the hardware. Godot could potentially leverage these extra queues (e.g. dedicated compute queues), to be more explicit about what work should be executed in parallel. Finding these paths that could be executed in parallel in the graph would require some elaborate detection of the dependencies between the commands, the possible paths that are independent and where the synchronization points need to be placed.
```
JohnVeness:
```suggestion
Pavlo Muratov's [article](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) that was used as the main inspiration behind this change contains a very interesting proposal in how to submit and synchronize GPU work across the multiple queues exposed by the hardware. Godot could potentially leverage these extra queues (e.g. dedicated compute queues), to be more explicit about what work should be executed in parallel. Finding paths that could be executed in parallel in the graph would require some elaborate detection of the dependencies between the commands, the possible paths that are independent and where the synchronization points need to be placed.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+| ![](/storage/blog/acyclic-graph/msaa-resolve-manual.webp) |
+|:--:| 
+| *It's not possible for Godot to know if something will draw again to the target, so it must resolve the result manually when it needs to sample the resource.* |
+
+With how flexible of an engine Godot is, determining where this step should go can be very tricky: the operation should be placed only when it's absolutely necessary or as part of the last render pass that draws to the MSAA texture. This is an area that the graph could aim to resolve automatically by simplifying the implementation of MSAA in the the renderers and lead to further performance improvements. Reducing the amount of resolve operations to the minimum and the bandwidth required for MSAA could be very beneficial for the Mobile renderer in particular.
```
JohnVeness:
```suggestion
With how flexible of an engine Godot is, determining where this step should go can be very tricky: the operation should be placed only when it's absolutely necessary or as part of the last render pass that draws to the MSAA texture. This is an area that the graph could aim to resolve automatically by simplifying the implementation of MSAA in the renderers and lead to further performance improvements. Reducing the amount of resolve operations to the minimum and the bandwidth required for MSAA could be very beneficial for the Mobile renderer in particular.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+| ![](/storage/blog/acyclic-graph/msaa-resolve-solution.webp) |
+|:--:| 
+| *Since the graph can detect if the render pass is the last one in the frame, it can automatically insert a resolve in the render pass, saving lots of bandwidth in the process.* |
+
+### Graph visualization
+
+While this wouldn't provide a direct benefit to end users, building a visualizer for the graph can help the Godot developers have a clear overview of the rendering pipeline of a given frame and identify bottlenecks more easily. During development, a few compute passes were identified that weren't being parallelized correctly due to implementation errors. For example, GPU particle systems were binding an unused buffer for write operations even if they never wrote to it, which led to the commands being identified as being dependent of each other due to having to synchronize with the potential "write" performed by the previous system.
```
JohnVeness:
```suggestion
While this wouldn't provide a direct benefit to end users, building a visualizer for the graph could help the Godot developers have a clear overview of the rendering pipeline of a given frame and identify bottlenecks more easily. During development, a few compute passes were identified that weren't being parallelized correctly due to implementation errors. For example, GPU particle systems were binding an unused buffer for write operations even if they never wrote to it, which led to the commands being identified as being dependent of each other due to having to synchronize with the potential "write" performed by the previous system.
```
"Could" sounds better when describing something that hasn't been built yet.

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+|:--:| 
+| *After fixing the error by assigning each system their own buffers, the graph automatically reordered and executed the particle systems in parallel, leading to huge gains in performance.* |
+
+While a more obvious case like this one was identified since it did not meet expectations at first, there could be more subtle instances of this behavior yet to be found in the codebase that could be easily exposed by building better debugging tools.
+
+## Conclusions
+
+While the use of a directed acyclic graph for rendering is not a brand new technique, the approach used by Godot is quite novel in many ways. The simplicity of the API exposed by Rendering Device no longer comes at the cost of rendering performance: the developer is guaranteed they'll get efficient use of the GPU and it can only get better from here. As the engine aims to be as general purpose as possible and maintain its ease of use, a lot of alternatives had to be discarded until the right approach was found. This is a long term technical investment that will pay off by reducing the cost of maintenance and unlocking new strategies for optimization in the long run.
```
JohnVeness:
```suggestion
While the use of a directed acyclic graph for rendering is not a brand new technique, the approach used by Godot is quite novel in many ways. The simplicity of the API exposed by Rendering Device no longer comes at the cost of rendering performance: the developer is guaranteed they'll get efficient use of the GPU and it can only get better from here. As the engine aims to be as general purpose as possible and maintain its ease of use, a lot of alternatives had to be discarded until the right approach was found. This is a long-term technical investment that will pay off by reducing the cost of maintenance and unlocking new strategies for optimization in the long run. 
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+An automatic approach ensures that despite the project being open-source and modified by many different people, they no longer need complete awareness of how the entire rendering pipeline works to modify it. This will also be very beneficial when PRs like [rendering effects](https://github.com/godotengine/godot/pull/80214) are merged, which introduces the ability to add post-processing steps where the Godot renderer will be completely unaware of all dependencies the hook may require. The extra validation introduced by recording commands to the graph has also exposed existing implementation errors, resulting in either bug fixes or even performance improvements.
```
JohnVeness:
```suggestion
An automatic approach ensures that despite the project being open-source and modified by many different people, they no longer need complete awareness of how the entire rendering pipeline works to modify it. This will also be very beneficial when PRs like [rendering effects](https://github.com/godotengine/godot/pull/80214) are merged, which introduce the ability to add post-processing steps where the Godot renderer will be completely unaware of all dependencies the hook may require. The extra validation introduced by recording commands to the graph has also exposed existing implementation errors, resulting in either bug fixes or even performance improvements.
```
Also, I'm not sure what is meant by "An automatic approach".

JohnVeness (commented):
I've finished my pass through the article. Congratulations on writing a readable document that even a non-graphics expert like myself could follow :)

JohnVeness:
I wonder if the diagrams should be SVG or something else editable and machine-parseable? I haven't spotted any glaring errors in them (although I would personally prefer text like "Render pass" or "Unrelated commands" instead of "Render Pass" or "Unrelated Commands") but it still might be better for accessibility or for people who might want to translate them etc.

collections/_article/rendering-acyclic-graph.md
DarioSamo: By "an automatic approach" I want to specify that no manual work is required from the engine developers to make the acyclic graph work. It's important to note this as other engines require describing it manually. Do you have a suggestion on how to make that easier to understand?

collections/_article/rendering-acyclic-graph.md
DarioSamo: I agree with both of your suggestions there, it is the more correct meaning.

DarioSamo:
> I wonder if the diagrams should be SVG or something else editable and machine-parseable? I haven't spotted any glaring errors in them (although I would personally prefer text like "Render pass" or "Unrelated commands" instead of "Render Pass" or "Unrelated Commands") but it still might be better for accessibility or for people who might want to translate them etc.

I did most of them in diagrams.net which technically allows you to export SVG, but the animated one is a series of images compiled into a webp, since I don't know how to make an SVG animation like that or if diagrams.net lets you make one (I did it very manually). So I guess there's no way to make the entire article consistently SVG, but at least they're lightweight enough as webp.

Calinou:
SVG animation exists as SMIL which is [pretty well-supported across browsers](https://caniuse.com/svg-smil), but not a lot of software can export to SMIL.

I think the current animated WebP approach is fine. Videos are typically preferable to animated images, but if you need lossless compression and full chroma, animated images are more suited.

coppolaemilio:
@DarioSamo For the article cover image, any screenshots of a project you are working with could do. Feel free to send a few and we can make it work.

DarioSamo:
> @DarioSamo For the article cover image, any screenshots of a project you are working with could do. Feel free to send a few and we can make it work.

Sure, here's some screenshots of the scene by W4 that was used for benchmarking.

[NUKU_SCREENSHOTS.zip](https://github.com/godotengine/godot-website/files/14284839/NUKU_SCREENSHOTS.zip)

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+excerpt: "Say goodbye to placing barriers, hello to the new acyclic graph"
+categories: ["progress-report"]
+author: Darío Banini
+image: /storage/blog/covers/todo.webp
+date: 2024-02-14 9:00:00
+---
+
+*Darío, the author of this article, is a new face around this blog who was hired by **W4 Games** last summer to start contributing to the project. His work was kindly sponsored and donated to the Godot Engine project by **W4 Games**. You can find more of his contributions to the engine in his [GitHub profile](https://github.com/DarioSamo).*
+
+Since the introduction of Godot 4, [Rendering Device](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) has been the backbone of the Forward+ and Mobile renderers. Making APIs that are both easy to use and flexible enough to cover all the features users want in a game engine is a very difficult task. The high level of detail expected by APIs like Vulkan or Direct3D 12 compared to their predecessors is worthy of a few blog posts of its own. This article will try to keep the topic as brief as possible to focus on the problems solved within Godot instead.
```
Calinou: RenderingDevice is a class name, so it should be written as PascalCase:

```suggestion
Since the introduction of Godot 4, [RenderingDevice](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) has been the backbone of the Forward+ and Mobile renderers. Making APIs that are both easy to use and flexible enough to cover all the features users want in a game engine is a very difficult task. The high level of detail expected by APIs like Vulkan or Direct3D 12 compared to their predecessors is worthy of a few blog posts of its own. This article will try to keep the topic as brief as possible to focus on the problems solved within Godot instead.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+If you're interested in the details of how this was achieved, keep on reading.
+
+## Background
+
+Understanding the problem space is crucial to do a deep dive into the technical decisions that were made. Dealing with an existing codebase for a general engine that had thousands of lines written on top of an existing API imposes many restrictions on what sort of changes are allowed. Flaws introduced in the early stages of design can have long-term effects on the development of a big project. This was very much the case with some of the key decisions taken for the Rendering Device abstraction, its coupling to Vulkan and the negative results on the rest of the codebase. The good news is it's never too late to go back to the drawing board. [reduz](https://github.com/reduz) laid out the plans for this [redesign](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08) during the end of 2023 and the team got to work on how to make it a reality.
```
Calinou:
```suggestion
Understanding the problem space is crucial to do a deep dive into the technical decisions that were made. Dealing with an existing codebase for a general engine that had thousands of lines written on top of an existing API imposes many restrictions on what sort of changes are allowed. Flaws introduced in the early stages of design can have long-term effects on the development of a big project. This was very much the case with some of the key decisions taken for the RenderingDevice abstraction, its coupling to Vulkan and the negative results on the rest of the codebase. The good news is it's never too late to go back to the drawing board. [reduz](https://github.com/reduz) laid out the plans for this [redesign](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08) during the end of 2023 and the team got to work on how to make it a reality.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+No changes are expected from users whatsoever. If you're one of the few people who have written code that uses Rendering Device, there's no need to worry either: while the API has changed slightly, all of the methods just require less information than before. What level of performance improvement you see will very much depend on the contents of your project. For example, GPU particle systems will get massive improvements, while more standard scenes that use post-processing will see some frametime reduction in the ballpark of around 5% to 15%. On top of that, the Godot developers will have a much easier time improving the performance of the renderers in the near future.
```
Calinou:
```suggestion
No changes are expected from users whatsoever. If you're one of the few people who have written code that uses RenderingDevice, there's no need to worry either: while the API has changed slightly, all of the methods just require less information than before. What level of performance improvement you see will very much depend on the contents of your project. For example, GPU particle systems will get massive improvements, while more standard scenes that use post-processing will see some frametime reduction in the ballpark of around 5% to 15%. On top of that, the Godot developers will have a much easier time improving the performance of the renderers in the near future.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+If you're interested in the details of how this was achieved, keep on reading.
+
+## Background
+
+Understanding the problem space is crucial to do a deep dive into the technical decisions that were made. Dealing with an existing codebase for a general engine that had thousands of lines written on top of an existing API imposes many restrictions on what sort of changes are allowed. Flaws introduced in the early stages of design can have long-term effects on the development of a big project. This was very much the case with some of the key decisions taken for the Rendering Device abstraction, its coupling to Vulkan and the negative results on the rest of the codebase. The good news is it's never too late to go back to the drawing board. [reduz](https://github.com/reduz) laid out the plans for this [redesign](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08) during the end of 2023 and the team got to work on how to make it a reality.
+
+### Vulkan
+
+There's a few Vulkan concepts that must be understood before digging into why Rendering Device was designed the way it was for Godot 4 and the difficulties it encountered along the way.
```
Calinou:
```suggestion
There's a few Vulkan concepts that must be understood before digging into why RenderingDevice was designed the way it was for Godot 4 and the difficulties it encountered along the way.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+| ![](/storage/blog/acyclic-graph/barriers.webp) |
+|:--:| 
+| *A barrier can establish a dependency between commands by specifying a synchronization scope.* |
+
+Beginners will have a hard time understanding this mechanism, but even experts or hardware vendors are not safe from making mistakes. A run of the [Vulkan Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers) in *synchronization* mode will reveal multiple issues on most Godot projects or even commercial games available in the market. These are among the most frustrating issues to understand as they won't even appear consistently depending on hardware vendor or hardware speed. Since eliminating these problems was one of the main goals behind the introduction of the acyclic graph, extensive use was made of these validation tools to ensure that no synchronization errors remained.
+
+### Rendering Device
```
Calinou:
```suggestion
### RenderingDevice
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+[Rendering Device is the abstraction](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) on which the *Forward+* and *Mobile* renderers in Godot were built. It exposes an interface that provides a reasonable level of control over the GPU in a way that is not as fine-grained as Vulkan or D3D12 but is also not as "stateful" as OpenGL. More of the rendering state needs to be defined in advance, and the chances of causing an error by leaking state from one previous command to the next are significantly reduced.
```
Calinou:
```suggestion
[RenderingDevice is the abstraction](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html) on which the *Forward+* and *Mobile* renderers in Godot were built. It exposes an interface that provides a reasonable level of control over the GPU in a way that is not as fine-grained as Vulkan or D3D12 but is also not as "stateful" as OpenGL. More of the rendering state needs to be defined in advance, and the chances of causing an error by leaking state from one previous command to the next are significantly reduced.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+While Rendering Device attempted to hide many of the difficulties that Vulkan introduced, there are a few that slipped through that made writing rendering code for Godot 4 more difficult than anticipated.
+
+#### Draw list actions
+
+Just as draw lists were naturally mapped to render passes, so too was the fact that the start and end action had to be specified for the textures associated with them. These actions cover operations such as loading the contents of a texture, discarding it or clearing it with a specific color. There's plenty of performance to be gained by choosing the optimal option in each render pass. However, due to some internal design decisions, the introduction of a "CONTINUE" action was made that doesn't actually exist in Vulkan. This action essentially meant "the previous draw list is compatible and the texture should not be transitioned to a different state". For an engine like Godot, which must provide a lot of **optional** post-processing effects and drawing layers, this proved to be a *maintenance nightmare*, ending in a ton of branching in the code that was not very easy to follow and prone to mistakes.
```
Calinou:
```suggestion
Just as draw lists were naturally mapped to render passes, so too was the fact that the start and end action had to be specified for the textures associated with them. These actions cover operations such as loading the contents of a texture, discarding it or clearing it with a specific color. There's plenty of performance to be gained by choosing the optimal option in each render pass. However, due to some internal design decisions, the introduction of a `CONTINUE` action was made that doesn't actually exist in Vulkan. This action essentially meant "the previous draw list is compatible and the texture should not be transitioned to a different state". For an engine like Godot, which must provide a lot of **optional** post-processing effects and drawing layers, this proved to be a *maintenance nightmare*, ending in a ton of branching in the code that was not very easy to follow and prone to mistakes.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### Barrier mask
+
+While Rendering Device does not provide explicit access to barriers, a lot of methods would allow specifying a bit mask to define which stages of the pipeline the command must synchronize with. However, this was probably one of the most misused masks in the existing rendering code as its implementation wasn't even consistent in most methods. In one instance, a variable that wasn't a bitmask at all was accidentally casted as one and passed to a rendering method.
```
Calinou:
```suggestion
While RenderingDevice does not provide explicit access to barriers, a lot of methods would allow specifying a bit mask to define which stages of the pipeline the command must synchronize with. However, this was probably one of the most misused masks in the existing rendering code as its implementation wasn't even consistent in most methods. In one instance, a variable that wasn't a bitmask at all was accidentally casted as one and passed to a rendering method.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+By default, most commands will just specify that all work that comes afterwards must be executed after the current command has finished. As expected, this results in Godot issuing an extreme amount of barriers "just in case" as it has no knowledge of what might be requested afterwards. Godot's strategy essentially boils down to issuing barriers to make sure no future work executes before the current command finishes, unless the programmer using Rendering Device specifies the barrier mask. Again, that means the developer must have exact knowledge of everything that comes next, and in turn, caused another round of maintenance issues. Therefore, this feature is rarely used and the default behavior is preferred, resulting in lower performance.
```
Calinou:
```suggestion
By default, most commands will just specify that all work that comes afterwards must be executed after the current command has finished. As expected, this results in Godot issuing an extreme amount of barriers "just in case" as it has no knowledge of what might be requested afterwards. Godot's strategy essentially boils down to issuing barriers to make sure no future work executes before the current command finishes, unless the programmer using RenderingDevice specifies the barrier mask. Again, that means the developer must have exact knowledge of everything that comes next, and in turn, caused another round of maintenance issues. Therefore, this feature is rarely used and the default behavior is preferred, resulting in lower performance.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+The initial redesign was laid out in *reduz*'s [draft](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08), which was largely inspired by [Pavlo Muratov's "Organizing GPU Work with Directed Acyclic Graphs"](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) and showed the possibility of how the concept could be applied to Godot's existing design. Not everything stated in the document made it into the final version: in fact, the changes to Rendering Device were much less severe than initially indicated and the interface remained largely compatible. While the article that was used as inspiration includes a very detailed algorithm in how to implement multi-queue submission by using the graph, the team made the decision to cut this idea short and stick to a single command queue to begin with, as the difficulty of the task would come from building the graph automatically and would already take a significant amount of development time.
```
Calinou:
```suggestion
The initial redesign was laid out in *reduz*'s [draft](https://gist.github.com/reduz/980b9b2547d57e6a915b2bb7e1e76e08), which was largely inspired by [Pavlo Muratov's "Organizing GPU Work with Directed Acyclic Graphs"](https://levelup.gitconnected.com/organizing-gpu-work-with-directed-acyclic-graphs-f3fd5f2c2af3) and showed the possibility of how the concept could be applied to Godot's existing design. Not everything stated in the document made it into the final version: in fact, the changes to RenderingDevice were much less severe than initially indicated and the interface remained largely compatible. While the article that was used as inspiration includes a very detailed algorithm in how to implement multi-queue submission by using the graph, the team made the decision to cut this idea short and stick to a single command queue to begin with, as the difficulty of the task would come from building the graph automatically and would already take a significant amount of development time.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Acyclic graph
+
+The unique aspect of the implemented graph is that its construction is completely invisible to the programmer using Rendering Device. Commands requested from the class are logged internally and each command maintains an adjacency list that is updated as new dependencies are detected. Since these adjacencies only work one way and older commands cannot depend on future commands, it is virtually impossible for cyclic dependencies to form (hence the "acyclic" part of the graph). While a graph can be constructed in many ways, a list of vertices and an adjacency list are sufficient. Render commands play the role of vertices, and commands store the indices of their adjacent commands.
```
Calinou:
```suggestion
The unique aspect of the implemented graph is that its construction is completely invisible to the programmer using RenderingDevice. Commands requested from the class are logged internally and each command maintains an adjacency list that is updated as new dependencies are detected. Since these adjacencies only work one way and older commands cannot depend on future commands, it is virtually impossible for cyclic dependencies to form (hence the "acyclic" part of the graph). While a graph can be constructed in many ways, a list of vertices and an adjacency list are sufficient. Render commands play the role of vertices, and commands store the indices of their adjacent commands.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+One important detail that resulted in frametime reductions during this step was to take into account the type of command as a sorting factor: grouping together operations based on whether they were related to copying data, drawing or compute dispatches provided some noticeable increases in performance. While the exact reason behind this has not been determined, it seems likely that GPUs prefer to change the type of pipeline they need to use within a command buffer as little as possible.
+
+While the concept of using a structure like a graph and using sorting algorithms might sound like the most daunting part of the task due to the level of abstraction involved, it is the dependency tracking and adjacency list detection during command recording where most problems arise. The relationships shown in the diagrams above were not specified by the programmer using Rendering Device: they must be detected automatically based on how the resources were used in the frame, and this turned out to be no small task due to some particular details of how Godot works.
```
Calinou:
```suggestion
While the concept of using a structure like a graph and using sorting algorithms might sound like the most daunting part of the task due to the level of abstraction involved, it is the dependency tracking and adjacency list detection during command recording where most problems arise. The relationships shown in the diagrams above were not specified by the programmer using RenderingDevice: they must be detected automatically based on how the resources were used in the frame, and this turned out to be no small task due to some particular details of how Godot works.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Resource tracking
+
+The resources used by Rendering Device in Godot are buffers or textures. While these are separate objects at the lower level depending on the API being used, the graph considers them both as one to share much of the logic during implementation. However, a distinction will be made later when texture slices are introduced, which is something Godot uses quite a bit in various parts of its rendering code. Textures also have the additional requirement that they need to make layout transitions to be ready for use in different commands, while buffers don't need to do this at all.
```
Calinou:
```suggestion
The resources used by RenderingDevice in Godot are buffers or textures. While these are separate objects at the lower level depending on the API being used, the graph considers them both as one to share much of the logic during implementation. However, a distinction will be made later when texture slices are introduced, which is something Godot uses quite a bit in various parts of its rendering code. Textures also have the additional requirement that they need to make layout transitions to be ready for use in different commands, while buffers don't need to do this at all.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Dependency tracking
+
+Since the graph construction is automatic and there's no input from the programmer in how the adjacency lists of the graph must be built, it's up to Rendering Device to use the resource trackers to figure out the dependencies between commands. While the final implementation is a bit more complex due to the introduction of texture "slices" (which is covered in another chapter), the main idea behind the algorithm is pretty straightforward.
```
Calinou:
```suggestion
Since the graph construction is automatic and there's no input from the programmer in how the adjacency lists of the graph must be built, it's up to RenderingDevice to use the resource trackers to figure out the dependencies between commands. While the final implementation is a bit more complex due to the introduction of texture "slices" (which is covered in another chapter), the main idea behind the algorithm is pretty straightforward.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+This usually doesn't pose any additional trouble if the engine sticks to using a texture during commands in only one particular way, but the problem does not stop there: Godot can and will use different parts of the same texture for completely different purposes, even within the same command. This is possible because [Rendering Device can create "shared" textures from slices](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html#class-renderingdevice-method-texture-create-shared-from-slice). While the original texture may only have one resource tracker, all shared slices are considered different textures with their own resource trackers that can be referenced independently in commands. A very common use case is mipmap creation: a lower level mipmap of the texture can be set as the render target, while a higher resolution level is used for sampling, effectively creating the chain of mipmaps from the texture's highest quality level.
```
Calinou: "Very" should be used sparingly; I've already seen it a lot of time this article.

```suggestion
This usually doesn't pose any additional trouble if the engine sticks to using a texture during commands in only one particular way, but the problem does not stop there: Godot can and will use different parts of the same texture for completely different purposes, even within the same command. This is possible because [RenderingDevice can create "shared" textures from slices](https://docs.godotengine.org/en/stable/classes/class_renderingdevice.html#class-renderingdevice-method-texture-create-shared-from-slice). While the original texture may only have one resource tracker, all shared slices are considered different textures with their own resource trackers that can be referenced independently in commands. A common use case is mipmap creation: a lower level mipmap of the texture can be set as the render target, while a higher resolution level is used for sampling, effectively creating the chain of mipmaps from the texture's highest quality level.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+| ![](/storage/blog/acyclic-graph/texture-slices.webp) |
+|:--:| 
+| *The anatomy of a cubemap texture that uses both mipmaps and array layers. Godot can create slices of a texture that reference only a range of its subresources.* |
+
+Tracking texture slices effectively required the implementation of a set of strict rules to verify that the programmer using the Rendering Device does not perform operations with undefined behavior.
```
Calinou:
```suggestion
Tracking texture slices effectively required the implementation of a set of strict rules to verify that the programmer using the RenderingDevice does not perform operations with undefined behavior.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+While these rules do not guarantee an optimal solution, it is not common for Godot to perform these operations on every frame, and the potential linearization of the commands is an acceptable performance hit to keep tracking as simple as possible. One point in particular that was discussed a lot was whether to resort to tracking usage individually per mipmap level and array layer, but such a decision would result in a system that wouldn't scale at all in the long run if large texture arrays with multiple mipmaps are used.
+
+## Results
+
+While this is something that is *not* usually measured, it's worth remembering one of the main reasons behind the graph was to simplify the maintenance of the rendering code for the team. The problems identified at the start clearly meant there was a lot of code that would get deleted from the rendering pipeline. While the PR might not have ended up in the net negative due to a lot of code being isolated in the graph's class along with debugging utilities, around ~2500 lines of code were removed from the implementation of Rendering Device, the Forward+ renderer and the Mobile renderer.
```
Calinou:
```suggestion
While this is something that is *not* usually measured, it's worth remembering one of the main reasons behind the graph was to simplify the maintenance of the rendering code for the team. The problems identified at the start clearly meant there was a lot of code that would get deleted from the rendering pipeline. While the PR might not have ended up in the net negative due to a lot of code being isolated in the graph's class along with debugging utilities, around ~2,500 lines of code were removed from the implementation of RenderingDevice, the Forward+ renderer and the Mobile renderer.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+The other major benefit is that a lot of hard to identify bugs that were caused due to synchronization bugs are now potentially fixed. As pointed out before, the [MSAA with SSAO issue](https://github.com/godotengine/godot/issues/61415) was resolved as a side-effect. Problems like this one proved to be extremely hard to fix as Godot developers would require the particular set of hardware and the right scenarios for the defects to trigger. With the graph sufficiently complete and unable to cause these synchronization problems, the possibility of introducing errors of this kind disappears.
```
Calinou: https://english.stackexchange.com/questions/364613/side-effects-or-side-effects

```suggestion
The other major benefit is that a lot of hard to identify bugs that were caused due to synchronization bugs are now potentially fixed. As pointed out before, the [MSAA with SSAO issue](https://github.com/godotengine/godot/issues/61415) was resolved as a side effect. Problems like this one proved to be extremely hard to fix as Godot developers would require the particular set of hardware and the right scenarios for the defects to trigger. With the graph sufficiently complete and unable to cause these synchronization problems, the possibility of introducing errors of this kind disappears.
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+|:--:| 
+| *After fixing the error by assigning each system their own buffers, the graph automatically reordered and executed the particle systems in parallel, leading to huge gains in performance.* |
+
+While a more obvious case like this one was identified since it did not meet expectations at first, there could be more subtle instances of this behavior yet to be found in the codebase that could be easily exposed by building better debugging tools.
+
+## Conclusions
+
+While the use of a directed acyclic graph for rendering is not a brand new technique, the approach used by Godot is quite novel in many ways. The simplicity of the API exposed by Rendering Device no longer comes at the cost of rendering performance: the developer is guaranteed they'll get efficient use of the GPU and it can only get better from here. As the engine aims to be as general purpose as possible and maintain its ease of use, a lot of alternatives had to be discarded until the right approach was found. This is a long-term technical investment that will pay off by reducing the cost of maintenance and unlocking new strategies for optimization in the long run.
```
Calinou:
```suggestion
While the use of a directed acyclic graph for rendering is not a brand new technique, the approach used by Godot is quite novel in many ways. The simplicity of the API exposed by RenderingDevice no longer comes at the cost of rendering performance: the developer is guaranteed they'll get efficient use of the GPU and it can only get better from here. As the engine aims to be as general purpose as possible and maintain its ease of use, a lot of alternatives had to be discarded until the right approach was found. This is a long-term technical investment that will pay off by reducing the cost of maintenance and unlocking new strategies for optimization in the long run. 
```

collections/_article/rendering-acyclic-graph.md
```diff
@@ -0,0 +1,277 @@
...
+
+An automatic approach ensures that despite the project being open-source and modified by many different people, they no longer need complete awareness of how the entire rendering pipeline works to modify it. This will also be very beneficial when PRs like [rendering effects](https://github.com/godotengine/godot/pull/80214) are merged, which introduce the ability to add post-processing steps where the Godot renderer will be completely unaware of all dependencies the hook may require. The extra validation introduced by recording commands to the graph has also exposed existing implementation errors, resulting in either bug fixes or even performance improvements.
```
Calinou:
```suggestion
An automatic approach ensures that despite the project being open source and modified by many different people, they no longer need complete awareness of how the entire rendering pipeline works to modify it. This will also be very beneficial when PRs like [rendering effects](https://github.com/godotengine/godot/pull/80214) are merged, which introduce the ability to add post-processing steps where the Godot renderer will be completely unaware of all dependencies the hook may require. The extra validation introduced by recording commands to the graph has also exposed existing implementation errors, resulting in either bug fixes or even performance improvements.
```

Calinou (changes requested):
Great article! I've only found some stylistic issues.

PR #805, opened 2024-02-16
Inline all svg files to reduce network requests

MichaelJRM:
I've made a simple plugin to help inline the svg files and also minified all of the relevant ones

Slightly improves performance


coppolaemilio:
I don't think inlining svg files is the way to go. While it can be a good optimization for the amount of requests, it makes them hard to be replaced with other formats in the future. I think it is better to treat image elements as image elements and not as extra markup. This PR would also add another dependency that might require for us to maintain, and I would prefer that we keep it as simple as possible. Hope it makes sense, thanks!

PR #806, opened 2024-02-16
Optimize home page

MichaelJRM:
- Updated tobbi & highlight to the most recent version
- Used the custom highlight download tool to only pick the languages that are currently in use thus reducing the file size
- Removed tobbi & highlight from all pages that do not use it
- Created small plugin to fetch CSS asynchronously
- Fixed header 

| Before  | After |
| before  | after |


Calinou:
> Created small plugin to fetch CSS asynchronously

I'm concerned this could introduce FOUC (flash of unstyled content). If the CSS isn't available yet, the page will be displayed as raw HTML.

The rest looks good to me :slightly_smiling_face:

pages/community/user-groups.html
```diff
@@ -202,3 +214,5 @@ <h3 class="title">{{ country[0] }}</h3>
 	</div>
 </main>
 {% include footer.html %}
+
+{% preload_async_css href="/assets/leaflet/leaflet.css" %}
```
Calinou: Add missing newline at end of file:

```suggestion
{% preload_async_css href="/assets/leaflet/leaflet.css" %}

```

MichaelJRM:
Yes I somehow didn't realize that before, It was only supposed to be used with CSS that's not critical :D

MichaelJRM:
Fixed a little bug with the fund button, I think I'm done here 🙂

_includes/footer.html
```diff
@@ -47,70 +47,57 @@
...
 		});
 
 		// Update any download link to point to identified user's OS.
 		const links = document.querySelectorAll('.set-os-download-url');
 		for (let i = 0; i < links.length; i++) {
 			const link = links[i];
-			let link_slug = 'download';
-			if ('version' in link.dataset && link.dataset['version'] === '3') {
-				link_slug = 'download/3.x';
+			let lingSlug = 'download';
```
Calinou: `ling`, shouldn't it be `link`?

Calinou:
Changes in this PR cause the homepage font size to change ever so slightly, resulting in line  wrapping being broken for the tagline:

Before | After *(this PR)*
-|-
image | Screenshot_20240221_190614

There's already `&nbsp;` between "waiting" and "for", so I don't know what's the cause of this.

Font selection is also different, causing all fonts to be slightly larger on my Linux setup:

Before | After *(this PR)*
-|-
image | image

Search icon alignment is broken on the Blog page:

Before | After *(this PR)*
-|-
image | image

MichaelJRM:
Fixed all but: "Font selection is also different, causing all fonts to be slightly larger on my Linux setup:"
Could It be due to the added [system-ui](https://drafts.csswg.org/css-fonts-4/#system-ui-def) font?  
What font do you have in your system?

`--default-font-family:  system-ui, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;`

MichaelJRM:
I've tested on mac and windows and the font looks the same, I think the issue was that before this pr it wasn't picking your system font.

_includes/footer.html
MichaelJRM: Good catch, thanks 😬

coppolaemilio:

PR #807, opened 2024-02-16
Optimize features page

MichaelJRM:
- Optimized the images carrousel

**This pr depends on #806**

| Before  | After |
|  before | after
| before| after


JohnVeness:
Out of interest, how do you obtain those performance measurement screenshots?

MichaelJRM:
Chromium based browser -> F12 -> Lighthouse
Always with a private window and cache disabled so it doesn't interfere with the results


lighthouse

Calinou (approved):
Changes specific to the Features page work and look good to me.

coppolaemilio:

PR #811, opened 2024-02-19
Optimize showcase page

MichaelJRM:
- Further compressed showreel images 

**This pr depends on #806**

| Before  | After |
| before | after |


Calinou:
Buttons are no longer on a single line with this PR, which looks odd on desktop as they're very wide:

Before | After *(this PR)*
-|-
image | image

The spacing between buttons on mobile makes them look better than before this PR though.

MichaelJRM:
Yes this is a question I had, before this pr the left section didn't respect the 50% width limit thus the buttons stay in a single line, after this pr the limit is respected and they fall into a column.

Should it stay like before or like now but instead of full width make them have the width of the largest element?

coppolaemilio:

PR #812, opened 2024-02-20
Update teams page

mhilbrunner:


raulsntos:
The scripting team section links to the `#scripting` channel which no longer exists because it was renamed `#dotnet`.
And the C# team has no link but should be linked to the `#dotnet` channel.
Also, can we rename the team `.NET` or `.NET/C#`?

mhilbrunner:
Renamed to `.NET`, linking to `#dotnet` and removed the reference to the no longer existing `#scripting` channel. Thanks!

pages/teams.html
```diff
@@ -167,8 +165,7 @@ <h4 class="title" id="editor">Editor • <a href="https://chat.godotengine.org/c
 	</div>
 
 	<div class="teams-team-section">
-		<h4 class="title" id="scripting">Scripting • <a href="https://chat.godotengine.org/channel/scripting">#scripting</a>
-		</h4>
+		<h4 class="title" id="scripting">Scripting</h4>
```
adamscott, line 168: Oops, it doesn't seem that the change to ".NET" was committed.

mhilbrunner: Should be below those lines, if I'm not missing anything. I don't think making the whole scripting section link to the #dotnet channel makes sense.

coppolaemilio (approved):

pages/teams.html
adamscott: Oh, right! Sorry!

adamscott: Makes me think that we should maybe just remove the scripting umbrella team all-together, especially if dotnet is going to be ported to GDExtension.

mhilbrunner: Yeah maybe, guess we can do followup work then though :) And no worries

coppolaemilio:
Thanks!

PR #821, opened 2024-03-14
Add post for RC2s of 4.2.2 and 4.1.4

clayjohn:
I am still missing a picture for the article, but I have everything else done


akien-mga (approved):
Should be good to go!

Issue #831, opened 2024-04-14
Homepage hero has horizontal scrollbar on small mobiles


snoopdouglas:
I think there's just been a copy change on the homepage, which on smaller mobiles causes a horizontal scrollbar to display on the hero of the homepage (because the text can't break). See the below video.

Can probably submit a PR for this myself but am currently AFK, will attempt this evening.

https://github.com/godotengine/godot-website/assets/4892574/7c97fafa-a0c7-4979-905c-ab2a575f8f18

snoopdouglas:
Note that this was originally observed on my ZenFone 9.

snoopdouglas:
Fixed by #833.

PR #833, opened 2024-04-14
Reduce homepage hero font size on mobile

snoopdouglas:
As per #831, reduces the font size of the hero header on mobile.

See below for iPhone SE (pretty small) and iPad (pretty large) viewports; I wonder if iPad is now too small and we need another breakpoint?

Screenshot from 2024-04-14 18-56-55

Screenshot from 2024-04-14 18-57-32

## Where this breaks

At 302px (on Firefox), this breaks and we get #831 again - but this is far tinier than the majority of phones now.

Screenshot from 2024-04-14 18-59-20

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -1258,7 +1258,7 @@ section.hero .copy p {
     margin: 0 auto;
   }
   section.hero .copy h1 {
-    font-size: 50px;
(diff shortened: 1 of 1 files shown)


coppolaemilio:
Thanks!

PR #836, opened 2024-04-16
Add blog post about the collaboration with Google and The Forge

clayjohn:
The date is arbitrary, I'm happy to post this whenever is convenient. 

Like the last post, this one has gone through internal review with Google, so we shouldn't make any major changes without consulting them first.

PR #838, opened 2024-04-19
Update cover on "collaboration-with-google-forge"

LuisCarli:
Update cover for article "collaboration-with-google-forge".


clayjohn:
Thank you very much for preparing this! I incorporated the image into https://github.com/godotengine/godot-website/pull/836 directly so it will go live with the article

PR #840, opened 2024-05-02
Make hero shorter and update call to actions

coppolaemilio:
# Current version
image

The current version of the hero is too tall. It includes 3 different prominent buttons to click and it hides the news under the fold. 

# This PR
image

This PR makes the overall height of the hero section to be shorter, and gives less weight to the Godot 3 download. 
The "learn more" button goes to the same place as the nav's "features" button, so it really doesn't make sense to have it so accessible there. 
It also adds a new link to the archive in case you need to download another version of the engine. This is something similar to what we or Blender have in the download page.

Supersedes https://github.com/godotengine/godot-website/pull/769

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -23,7 +23,7 @@
 
 			<ul class="right">
 				<li><a href="/download/windows" class="set-os-download-url">Download</a></li>
-				<li><a href="https://docs.godotengine.org">Learn</a></li>
+				<li><a href="https://docs.godotengine.org">Docs</a></li>
 				<li><a href="https://docs.godotengine.org/en/stable/community/contributing/index.html">Contribute</a></li>
 				{% comment %}
 				We add the control character at the end to force the heart to render as text and not as an emoji.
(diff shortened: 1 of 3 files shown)


winston-yallow (approved):

akien-mga:
Looks good to me overall, but the line for Godot 3 is really subtle. On that hero at least it's super easy to miss.

I agree with making it less prominent (and remove the misleading LTS terminology), but it should probably still be slightly clearer.

JohnVeness:
Proposal looks good, except I would make the width of the "Develop your 2D & 3D games..." paragraph the same as the header above. At the moment, it wraps too early, which looks messy.

Also worth checking how it looks on mobile?

coppolaemilio:
> Looks good to me overall, but the line for Godot 3 is really subtle. On that hero at least it's super easy to miss.

When the page is full size (not in a screenshot) it is not as easy to miss, but even if they miss it, we already prompt people via all the different paths we have to download:

If you click on the Nav > Download, or the Download Latest CTA, you will see:

image


So that's 3 different ways to get to the download Godot 3 page

coppolaemilio:
@JohnVeness Mobile didn't change much:

image

For the paragraph, I agree, completely missed making the update there as well.

Should look like this now:
image

IdfbAn:
Why replace "Learn" with "Docs"?

PR #662, opened 2023-07-03
Add Lemmy community card

jsbeckr:
Due to RedditMigration a few have us have migrated to a Lemmy community over at programming.dev

I would like to add a card, so that other people know about it.

PR #841, opened 2024-05-03
Add Lemmy community card

Kissaki:
Replaces #662. Rebased and section moved. As per request at https://programming.dev/post/13548997 and [the last comment on that PR](https://github.com/godotengine/godot-website/pull/662#issuecomment-1936992479).

I made some additional adjustments to the previous state too that seemed appropriate to me:

* img png -> webp
* gradient background like the one next to it
* consistent formatting to the element next to it
* consistent asset name to the one next to it (different to other "icon" assets for official communities/platforms)


Kissaki:
Tagging @coppolaemilio

Kissaki:
I have not actually run this [yet]. _(I currently have no Jekyll installed/set up.)_

coppolaemilio:
Thanks for the PR. Seems to be good! the only issue I see is on mobile, where it doesn't properly go full with on those cards. Can't check the fix now, but shouldn't be that hard to figure out

image

Both cards should go full width here instead of sharing the same row.

Ategon:
Note the text should probably say programming.dev instead of lemmy as lemmy is just the backend and we aren't actually using it (switching off for sublinks eventually and have our own dev team). Also lots of other users participating from places like piefed, mastodon, iceshrimp, kbin, mbin, etc.

Ategon:
The image issues are caused by the text being larger and pushing the image up for space. Could be fixed by setting a min size but then it would break on some smaller screen sizes where the other images are resizing to fit. Easiest fix would be shortening the text. edit: actually seems like maybe the text extends the cards down for some of the other cards but isnt for some reason on the new one. I can install jekyll later and test around with it

Kissaki:
Looks like the dynamic layout in other sections or sections blocks in general are solved through layout breakpoints. I guess this section is probably missing one - given it had only one card before (+ other kind of blocks).

Kissaki:
Looks like the current styling is not robust against sizes beyond the limited content all other blocks conform to

image

Kissaki:
I can't even simply overflow hide (with text ellipsis) I think because it uses wider space because the image width is defined wider. I guess this needs bigger changes to the generic layout.

Calinou:
I think the only thing you could do for now is using a smaller font size for this particular heading.

Kissaki:
I followed the title font size resolution suggestion. The result seems acceptable, and it's the most reasonable [PR] finishable solution.

Kissaki:
@coppolaemilio can you please check again? For me, it breaks into a one-column layout below `548px` viewport width, and above that, the two columns look fine.

Kissaki:
(screenshots without the img)

image

image

coppolaemilio:
> Note the text should probably say programming.dev instead of lemmy as lemmy is just the backend and we aren't actually using it (switching off for sublinks eventually and have our own dev team). Also lots of other users participating from places like piefed, mastodon, iceshrimp, kbin, mbin, etc.

How strict are you with this? because on the site itself it says `join-lemmy.org` at the bottom, and lemmy has way more brand recognition about it than programming.dev. 
I think if people see lemmy they are more likely to expect a "reddit replacement" than if they see the domain name of a website they might not know.

coppolaemilio:
I would propose something like this:
image

Ategon:
@coppolaemilio Thats on one of the frontends which is currently the default since sublinks isnt production ready yet. Weve got 4 other browser frontends and a bunch of apps that can access the site though.

https://p.programming.dev
https://old.programming.dev
https://v.programming.dev
https://a.programming.dev

Just once sublinks is ready Lemmy is not going to be used apart from being an alternate frontend until something happens to break frontend api compatibility. I view it more as theres a group of reddit-like sites that can all talk to each other using a shared compability backend through Lemmy rather than Lemmy being its own entity and sites being subsets of that (since they really aren't, all Lemmy sites are very different with things like Beehaw giving a completely different experience to something like Hexbear. And other Lemmy sites also have their own Godot communities (much smaller than the P.D one but still exist and have some activity))

If you want the software to be used for sure could be Lemmy for now and then swap to Sublinks in the future once that goes through on the site but I do prefer the site itself. Could note in the description it uses Lemmy or something if you want the brand recognition from that.

Could also capitalize it to be Programming.Dev to be consistent with all the other cards

Kissaki:
> Could also capitalize it to be Programming.Dev to be consistent with all the other cards

I chose lowercase because that's how the instance names itself in the sidebar.

Kissaki:
Lemmy is a technology and an ecosystem / platform style.

@Ategon mentioned some of the diffusion / unclear borders and potential change in time. I think there's even more but I'll refrain from laying out more.

I think "community for discussion" is a fine description.

Lemmy may have some recognizability, but I think we get more problems with using it than gain. I assume most people probably won't know what Lemmy is anyway. And if they try to learn about it, it's a whole other ordeal of potential confusion.

Kissaki:
Maybe mentioning Fediverse primarily and potentially Lemmy secondarily may be more appropriate or useful. But personally, I think that's too technical and only telling for people who already know what they are.

coppolaemilio:
Regarding the name, we list mastodon as our social media platform for it, but the instance is at `gamedev.place`. I wouldn't make the card be `gamedev.place` because no one will know what they will get into, and no one will probably click on it. With this I feel a bit the same. I did know what Lemmy was, and every time there is a discussion about Reddit getting replaced by _something_ Lemmy comes up. So if I see an open source project that I like listing the places where the community hangs out, I would much rather see `Lemmy` than `instancedomain.here` as the card title. But I could be completely wrong and I don't want to miss-represent the community from programming.dev. I'm happy to call it whatever y'all prefer because it already seems active enough regardless of the name.

@Ategon let me know your preference and I can make the final edits to push this one live :)

Ategon:
Ah yeah the sidebar hasnt been updated in a bit so I guess it would still say programming.dev. Usually the new documents I write up have it as Programming.Dev like the legal docs or some of the other more recent supporting sites.

Preference is still having it as the site name since thats our brand and its whats getting supported as our brand in other spots + wont be changed (uppercase preferred to match the recent docs and seem more like a name so Programming.Dev)

coppolaemilio:
All right, made the latest adjustments and I'll merge now. Thanks!

PR #845, opened 2024-05-11
Remove showcase submission form

winston-yallow:
The form is currently not working. Long term we should probably find a proper solution, but for now a quick fix is to remove the link so that people are not misguided.


coppolaemilio:
thanks

PR #846, opened 2024-05-13
Add way to customize host and port for `build-and-serve.sh`

adamscott:
With this PR, we can customize the host and port for `build-and-serve.sh` utility script.

```console
# Sets the host to localhost instead of the default 0.0.0.0
SERVER_HOST=localhost ./build-and-serve.sh
# Sets the port to 4001 instead of the default 4000
SERVER_PORT=4001 ./build-and-serve.sh
# Sets both values
SERVER_HOST=localhost SERVER_PORT=4001 ./build-and-serve.h
```

### DIFF
--- a/build-and-serve.sh
+++ b/build-and-serve.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
 
+SERVER_HOST="${SERVER_HOST:-0.0.0.0}"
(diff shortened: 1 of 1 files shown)


coppolaemilio (approved):

coppolaemilio:
"${THANKS:-0.0.0.0}"

PR #848, opened 2024-05-15
Fix C# platform support table

raulsntos:
Mono requires .NET **7.0**, since we use `coreclr_create_delegate` which wasn't implemented in Mono before 7.0.
- See https://github.com/dotnet/runtime/pull/59962.

Thanks @paulloz for letting me know that the table was incorrect.


coppolaemilio:
thanks!

PR #850, opened 2024-05-18
Update URLs for Croatia

wingedadventurer:
Replace URLs for Croatia Godot-specific servers (which are getting deleted) with general-purpose gamedev servers.

----

We have decided to delete Godot-specific Discord server and Facebook website due to very low usage and demand.

I have replaced the URLs with general-purpose gamedev server/website that we have which has a ton more users. Discord server has a #engine-godot channel, and Facebook website is open for any discussions, including Godot.

In case of this PR not getting approved, I will update it and delete Croatia from the list as it won't have any Godot-specific servers/websites starting May 26th.


coppolaemilio:
We will eventually revamp this section, but I'm fine with giving an alternative if that one is going dark. Thanks for the PR!

PR #857, opened 2024-05-31
Avoid redirects in internal links by appending trailing slash

Calinou:
The web server redirects pages like `/blog` to `/blog/`, which slows down navigation a bit. This can be avoided by fully spelling out the URLs in internal links.

This also adds the download link OS matching to the 404 page, and fixes the asset library link in the footer when hosting the site locally (so it points to godotengine.org).

I've tested this locally and navigation still works as expected.


coppolaemilio (approved):
Thanks! I think we can merge this, but I also wonder if we can change the .htaccess file to make it server the correct file with the redirect

PR #863, opened 2024-06-02
Fix `minify` generation of `./_site/`

### LINKED ISSUE #860: Minifying step is not properly setup for the deployed version of the site.
The local version of the Jekyll site makes a good job of minifying assets and exporting them to the `/_site/` directory.
This works all right since when you are developing locally, the generated files all live at `/_site/`. 

The issue comes when building for deploying on the server, the minifying step is generating the ouput in `/_site/` means that instead of just replacing the ones at the root `/` it creates a duplicate at `/_site/`. You can see this by visiting  https://godotengine.org/_site/ which is actually a minified clone version of the website. 

The solution would be to make the necessary adjustments in the CI so that when building the site for release, we don't output the minified files in `/_site/`.

adamscott:
Needs to be tested, but works locally. (but I could reproduce locally the issue, so it should work)

Fixes #860

### DIFF
--- a/_plugins/minify-assets.rb
+++ b/_plugins/minify-assets.rb
@@ -1,5 +1,8 @@
 # Minify assets after each build
 Jekyll::Hooks.register :site, :post_write do
   puts "Minifying assets"
-  `minify -r -o _site/ _site/`
+  # Project path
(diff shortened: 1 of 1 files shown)


coppolaemilio:
Let's give it a go! thanks :)

PR #864, opened 2024-06-03
Fix relative path issue with `minify`

### LINKED ISSUE #860: Minifying step is not properly setup for the deployed version of the site.

adamscott:
The minify generation had the following error on the github action:
```
Minifying assets
ERROR: stat .//home/runner/work/godot-website/godot-website/_site: no such file or directory
                    done in 8.002 seconds.
```

This now uses ruby `pathname` library to get a relative path.

Follow-up of #863
Fixes #860

### DIFF
--- a/_plugins/minify-assets.rb
+++ b/_plugins/minify-assets.rb
@@ -1,8 +1,11 @@
+require 'pathname'
+
 # Minify assets after each build
 Jekyll::Hooks.register :site, :post_write do
   puts "Minifying assets"
   # Project path
-  site_path_from = File.join(Dir.pwd, "_site")
-  site_path_to = Dir.pwd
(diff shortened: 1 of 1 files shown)


winston-yallow (approved):

winston-yallow:
Thanks!

PR #866, opened 2024-06-04
Drop 'Blender ESCN exporter' from download pages

### LINKED ISSUE #713: "Godot 3 also supports glTF 2.0 and OBJ" on Download pages
On the various Download pages, there is the text "Godot 3 also supports glTF 2.0 and OBJ". I was wondering if we could drop the "3" from that now, assuming 4 still supports those?

Kissaki:
[Blender ESCN exporter](https://github.com/godotengine/godot-blender-exporter) is unmaintained and experimental with [51 open bug tickets][openbugs].

As discussed in #713, drop the prominent mention on the download pages.

Fixes #713

I haven't run it locally (no Ruby/Jekyll env), but I tested with DOM changes directly.

The CSS break point `768px` looks better now that it's two columns, so no need to change it despite now being two instead of three layout/card columns.

[openbugs]: https://github.com/godotengine/godot-blender-exporter/issues?q=label%3A%22bug%22+is%3Aopen+sort%3Aupdated-desc

### DIFF
--- a/_layouts/download-3.html
+++ b/_layouts/download-3.html
@@ -6,7 +6,7 @@ layout: default
 {% include header.html %}
 
 <link rel="stylesheet" href="/assets/css/transparent-nav.css">
-<link rel="stylesheet" href="/assets/css/download.css?2">
+<link rel="stylesheet" href="/assets/css/download.css?3">
 <style>
 	.digital-stores .digital-store-list {
 		grid-template-columns: 1fr 1fr;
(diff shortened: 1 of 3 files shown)


Calinou (approved):
This makes sense, glTF and FBX import have pretty much superseded it entirely.

ESCN handling code remains in the editor in case someone wants to write their own ESCN exporter, but I don't think there's much point in maintaining our own now.

coppolaemilio:
thanks!

PR #870, opened 2024-06-13
Add GodotCon 24 article

MurderVeggie:
Wrote the article, created the necessary files, but some links are still missing.


coppolaemilio (commented):
the authors/nat.png image is way too big, we usually go with smaller sizes, try to look at the others to compare. Also, webp is preferred but not required

MurderVeggie:
apologies, should've just renamed the already downsized file, forgot it was there

winston-yallow (approved):

PR #878, opened 2024-07-04
4.3 release page

adamscott:
> [!IMPORTANT]  
> **I need your help!** See the note below on how you can help me.

<details>
<summary><strong>Preview of the release page</strong></summary>

image

</details>

## Help needed
I need help to prepare demos for the media. I've done some already, but I need to fill all the media spots that say: "TODO: replace me".

If you plan to create a demo, please comment and say so. It will save some time for everyone.

### What I mean by demo
What I mean by demo is a simple project that I can unpack on my machine and capture the content, either with an image or by recording the content. I'll use my Macbook Pro as it has a nice resolution/high DPI.

### Demos missing
- [ ] (Outstanding) Interactive music support
- [x] (Outstanding) TilemapLayer node
- [ ] (2D) Parallax2D node
- [ ] (GDScript) GDScript autocompletion fixes
- [ ] (Rendering) Premultiplied alpha blending in 3D shaders
- [ ] (Web) splashscreen now showing up
- [ ] (XR) Composition layers to allow 2D viewports to be displayed at much higher quality
- [ ] (XR) Standardized and enhanced support for hand/finger tracking, body tracking, and face tracking


coppolaemilio:
# Style
The download section should be the first thing on the page, you can add the video and the rest of the features afterwards. Most people will come just for the download, and making them scroll should be optional. 

I think we can use the hero to highlight an image of a game, something like PVKK instead of having the logo two times.
While the monochrome style looks nice, it gets a bit confusing when comparing versions:
image

The padding in the title is inconsistent, and making the font too big requires for more space for it to not feel too crammed:
image

The donate button doesn't look like a button, and the background of the godot logos cutting don't look like there is a bug in the site. 
image


There isn't enough contrast in "accent" titles for the cards that have gradient backgrounds, the words "Layer up" are difficult to read:
image


I'm not sure if mobile was supposed to be in already or if that is something to be added in the future, because it looks incomplete in most of the cases. 


# Code
I would recommend to split the css into a .scss file (it can be in the same directory).

I would replace the grid-template areas with more generic column layouts and use those names instead of describing the content of them. 

```
grid-template-areas:
	/* Header */
	"presentation"
	"foundation"
	"cannot-wait"
	"spacer-after-cant-wait"
	/* Features: */
	"outstanding"
	/* Features alphabetically ordered: */
	"_2d"
	"animation"
	"dotnet" /* Should read as "C#". */
	"documentation"
	"display"
	"editor"
	"gdscript"
	"rendering"
	"web"
	"xr"
	/* Footer */
	"spacer-before-download"
	"download";
```

If we were going to reuse this for the base for a future page, we would need to start renaming all the template areas which will be a bit of a mess. So by using agnostic names like span-col-2.

I don't understand the need to make a shadow-root with this much style for each element: 
image
Everything that is currently implemented can be achieved with minimal css and standard html.
There is a lot of stuff being generated with Javascript without any real need, so I would try to instead of reinventing how to render elements in html I would stick to standards.

JohnVeness:
Looking at the preview image you kindly put above, I see a typo of "orignially". Also "open source game engine" should be "open-source game engine". Probably more but it's quite hard to read and edit the 4.3.html file, with so much stuff around the blocks of text.

Also, there seems to be a lot of hyphenation where I wouldn't expect it. It looks odd seeing words like "hundreds" or "platform" hyphenated.

pages/releases/4.3.html
```diff
@@ -0,0 +1,2337 @@
...
+			<!-- Presentation -->
+			<div id="presentation" class="release-section release-section--presentation">
+				<godot-release-box
+					type="bottom"
+				>
+					<div slot="blockquote">
+						Here is a video from one of our content-creators!
+					</div>
+					<div slot="media">
+						<iframe style="width: 100%; aspect-ratio: 16/9; height: min-content;" width="560" height="315" src="https://www.youtube.com/embed/bbatLr98fEY?si=LvzwlTzH9ZLRJzRK" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
```
clayjohn, line 1392: I assume this youtube video is a placeholder? If so, what do you intend to put here?

pages/releases/4.3.html
```diff
@@ -0,0 +1,2337 @@
...
+						<h2>Godot Foundation</h2>
+					</a>
+				</div>
+				<godot-release-box
+					class=""
+					inverted
+					type="right"
+				>
+					<div slot="blockquote">
+						Your <span class="highlight">donations</span> did this!
```
Calinou, line 1410: Does this fit within two lines? It sounds more natural to me.

```suggestion
						Your <span class="highlight">donations</span> made this possible!
```

Otherwise, "made it possible!" could perhaps fit.

Calinou:
I'll look into creating a premultiplied alpha blending demo (most likely a VFX/particles comparison with and without premultiplied alpha).

**Edit:** I made something, what do you think? It showcases how a premultiplied alpha material can exhibit both additive and mix blending at the same time (while the Mix/Add blend modes can only do one of those at once).

[blend_mode_premul_alpha.mp4.zip](https://github.com/user-attachments/files/16114305/blend_mode_premul_alpha.zip)

Project files: [test_3d_premul_alpha.zip](https://github.com/user-attachments/files/16114300/test_3d_premul_alpha.zip)

Also, what's the expected aspect ratio for the video?

pages/releases/4.3.html
adamscott: Oops, yes, it's a placeholder. We'll try to put a video from a content-creator there that is explaining the update on a video form. For an example, [see the Blender 4.1 page](https://www.blender.org/download/releases/4-1/).

adamscott:
> If we were going to reuse this for the base for a future page, we would need to start renaming all the template areas which will be a bit of a mess.

The grid system was used because it decouples the declaration from the presentation. We need to switch up the order last minute? No problem, we just have to switch it there instead of copy pasting large swaths of code.

If we were going to reuse this for the base for a future page, the categories wouldn't change much, so I'm not worried about this.

> I don't understand the need to make a shadow-root with this much style for each element:

This style inside the shadow-root is only repeated there for convenience. It isn't repeated download-wise, "declared once, used multiple times".

> I would try to instead of reinventing how to render elements in html I would stick to standards.

With all due respect, [this is standard](https://developer.mozilla.org/en-US/docs/Web/API/Web_components).

coppolaemilio:
> With all due respect, [this is standard](https://developer.mozilla.org/en-US/docs/Web/API/Web_components).

Well, if that is a standard we are truly lost in the web front ⚰️🪦

adamscott:
By the way, a new PR will be made, I'm creating a other version based on @coppolaemilio's comments.

PR #885, opened 2024-07-28
Update download page to add Windows arm64 binaries

adamscott:
This PR:
- adds Windows ARM64 links
  - supports 4.3.rc1 links
- renames x86_64 to x86-64
- ~~renames build links to more standard names for 4.3.rc2 and onwards~~
  - ~~Standard:~~
    - ~~linux.64: linux_x86-64.zip _(instead of linux.x86_64.zip)_~~
    - ~~linux.32: linux_x86-32.zip _(instead of linux.x86_32.zip)_~~
    - ~~linux.arm64: linux_arm64.zip _(instead of linux.arm64.zip)_~~
    - ~~linux.arm32: linux_arm32.zip _(instead of linux.arm32.zip)_~~
    - ~~macos.universal: macos_universal.zip _(instead of macos.universal.zip)_~~
    - ~~windows.64: windows_x86-64.zip _(instead of win64.exe.zip)_~~
    - ~~windows.32: windows_x86-32.zip _(instead of win32.exe.zip)_~~
    - ~~windows.arm64: windows_arm64.zip _(instead of windows_arm64.exe.zip)_~~
  - ~~Mono:~~
    - ~~macos.universal: mono_macos_universal.zip _(instead of mono_macos.universal.zip)_~~
    - ~~windows.64: mono_windows_x86-64.zip _(instead of mono_win64.zip)_~~
    - ~~windows.32: mono_windows_x86-32.zip _(instead of mono_win32.zip)_~~
- adds x86 and x86-64 tags
- does some reordering/cleanup in download_configs.yml
  - removed the weird space between extras and mono
  - put extras after mono
- removes "(x86_64 - Apple Silicon)" from macOS build name
- updates Android tags

~~Commit https://github.com/godotengine/godot-website/pull/885/commits/48fe40b4b3bf809c292fa62ffcf2ad71c4015aa0 renames mono to dotnet _(dotnet.windows.x86_64.zip instead of mono_windows_x86_64.zip)_. Easily revertable if it's not what we want.~~ _Removed because `dotnet` is reserved by the dotnet team for the module._

| Before | After |
| ---- | ---- |
| Capture d’écran, le 2024-07-29 à 11 05 51 | Capture d’écran, le 2024-07-29 à 11 09 14 |

### DIFF
--- a/_data/download_platforms.yml
+++ b/_data/download_platforms.yml
@@ -14,64 +14,73 @@
 
 - name: "android.apk"
   title: "Android"
-  caption: "Universal APK (ARM64 + ARMv7 + x86_64 + x86)"
+  caption: "Universal APK"
   tags:
     - APK download
-    - ARM64/v7
-    - x86 (64 & 32 bit)
+    - arm64
+    - arm32
+    - x86_64
+    - x86_32
 
 - name: "android.playstore"
   title: "Android"
-  caption: "Play Store Universal (ARM64 + ARMv7 + x86_64 + x86)"
+  caption: "Play Store Universal"
   tags:
     - Play Store
-    - ARM64/v7
-    - x86 (64 & 32 bit)
+    - arm64
+    - arm32
+    - x86_64
+    - x86_32
 
 - name: "linux.32"
   title: "Linux"
-  caption: "Standard (x86)"
+  caption: "Standard (x86_32)"
   tags:
+    - x86_32
     - 32 bit
 
 - name: "linux.64"
   title: "Linux"
   caption: "Standard (x86_64)"
   tags:
+    - x86_64
     - 64 bit
 
 - name: "linux.arm32"
   title: "Linux"
-  caption: "Standard (ARM32)"
+  caption: "Standard (arm32)"
   tags:
-    - ARM32
+    - arm32
     - 32 bit
 
 - name: "linux.arm64"
   title: "Linux"
-  caption: "Standard (ARM64)"
+  caption: "Standard (arm64)"
   tags:
-    - ARM64
+    - arm64
     - 64 bit
 
 - name: "linux_server.64"
   title: "Linux Server"
   caption: "Standard (x86_64)"
   tags:
+    - x86_64
     - 64 bit
 
 - name: "linux_server.headless.64"
   title: "Linux Server"
   caption: "Headless (x86_64)"
   tags:
+    - x86_64
     - 64 bit
     - Headless
 
 - name: "macos.universal"
   title: "macOS"
-  caption: "Universal (x86_64 + Apple Silicon)"
+  caption: "Standard"
   tags:
-    - Intel/Apple Silicon
+    - Apple Silicon
+    - x86_64
     - 64 bit
 
 - name: "web"
@@ -83,12 +92,21 @@
 
 - name: "windows.32"
   title: "Windows"
-  caption: "Standard (x86)"
+  caption: "Standard (x86_32)"
   tags:
+    - x86_32
     - 32 bit
 
 - name: "windows.64"
   title: "Windows"
   caption: "Standard (x86_64)"
   tags:
+    - x86_64
+    - 64 bit
+
+- name: "windows.arm64"
+  title: "Windows"
+  caption: "Standard (arm64)"
+  tags:
+    - arm64
     - 64 bit
--- a/_data/download_configs.yml
+++ b/_data/download_configs.yml
@@ -10,10 +10,8 @@ defaults:
       macos.universal: macos.universal.zip
       windows.64: win64.exe.zip
       windows.32: win32.exe.zip
+      windows.arm64: windows_arm64.exe.zip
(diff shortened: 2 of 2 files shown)


Calinou:
Do we have to display "Standard" and "Universal" next to all download names? I thought "Standard" was the opposite of ".NET" anyway…

The macOS downloads already list "(x86_64 + Apple Silicon)", which is why I don't think listing Universal is needed. If we keep listing Universal for it, then we could remove "(x86_64 + Apple Silicon)" (or even use the more technically correct term for it, "Universal **2**").

akien-mga (changes requested):
Haven't reviewed yet, but just marking that this shouldn't be merged without my review.

Renaming binaries here means change the way we do our official builds to match that, so I need to work on that in parallel.

adamscott:
I pushed a change so that RC1 and RC2 will use the same current file names.

mhilbrunner:
Consistency is great, and this is definitely an improvement I've wished for in that regard.

Still, this has a cost. I had to redo our automations/CI scripts 3 point releases ago for 4.0 at W4, now this will make us touch these again. Its not a huge thing, but is an annoyance, and I'm not 100% sure it is worth it to incur that to everyone for a bit more consistency. The nerd in me likes the improvement, but at some point we should stop breaking people's stuff constantly, because we can *always* improve stuff, and it does get tiresome eventually.

Long text to say *not sure if worth*.

Repiteo:
@mhilbrunner In a vacuum, I could see this point having more weight. However, given this is accompanied with adding a new link outright, it feels entirely within scope to make adjustments to other links.

akien-mga:
So thinking some more about it, I'm erring more on the side on not renaming the other binaries for now. I already had the concern that Max expressed, and I also lack time to do a deep review and changes to the naming scheme - and we're also super late in the release cycle, with RC3 in the oven and no time to beta-test those changes.

We could do a review and possibly renames for 4.4, or wait for the future release which will drop support for the legacy "mono" binaries, and rename everything at that point.

For now, I would suggest just adding the arm64 binaries with the current `_windows_arm64.zip` pattern.

This can be done in this PR, or in a new one if we want to keep this one as a draft for reference for potential future renames of all binaries.

And so the pattern should be valid for 4.3-rc1 onwards, including 4.3-stable and beyond.

> @mhilbrunner In a vacuum, I could see this point having more weight. However, given this is accompanied with adding a new link outright, it feels entirely within scope to make adjustments to other links.

IMO it's the other way around. Adding a new link doesn't break compat for user scripts. If they don't need Windows arm64 editor binaries, they don't need to do any changes. So breaking compat due to the addition of a platform that most users (for now) don't really care about would generate a lot of unnecessary friction.

_data/download_platforms.yml
```diff
@@ -14,30 +14,36 @@
...
   title: "Linux"
   caption: "Standard (x86)"
   tags:
+    - x86
     - 32 bit
 
 - name: "linux.64"
   title: "Linux"
-  caption: "Standard (x86_64)"
+  caption: "Standard (x86-64)"
```
akien-mga: We consistently use `x86_64` in Godot, AFAIK only macOS uses `x86-64` (which they now deprecated :upside_down_face: ), so I would undo that change and make the tag `x86_64`.

akien-mga: Actually I'd suggest harmonizing all arch names and tags to the ones we use in Godot:
- `x86_64`
- `x86_32`
- `arm64`
- `arm32`

_data/download_platforms.yml
```diff
@@ -14,64 +14,73 @@
...
 - name: "android.playstore"
   title: "Android"
-  caption: "Play Store Universal (ARM64 + ARMv7 + x86_64 + x86)"
+  caption: "Play Store Universal"
   tags:
     - Play Store
-    - ARM64/v7
-    - x86 (64 & 32 bit)
+    - arm64
+    - ARMv7
```
akien-mga:
```suggestion
    - arm32
```

_data/download_platforms.yml
```diff
@@ -14,64 +14,73 @@
...
 - name: "android.apk"
   title: "Android"
-  caption: "Universal APK (ARM64 + ARMv7 + x86_64 + x86)"
+  caption: "Universal APK"
   tags:
     - APK download
-    - ARM64/v7
-    - x86 (64 & 32 bit)
+    - arm64
+    - ARMv7
```
akien-mga:
```suggestion
    - arm32
```

akien-mga (approved):
Looks good to me!

akien-mga:
Thanks!

adamscott:
I'm having some issues with the PR, I'm debugging.

Capture d’écran, le 2024-08-08 à 10 27 26

mhilbrunner:
Yeah, I don't know how but I misread this as also changing the URLs of existing downloads for other platforms. May have been looking at different PR? Sorry for the confusion :)

PR #901, opened 2024-08-16
Improve .NET note on the main download page.

bruvzg:
A small update to the .NET note, now we have support for arm64 the same limitation applies to running x86_32 and x86_64 .NET apps on arm64, not only to 32-bit apps on 64-bit windows.


coppolaemilio (approved):

coppolaemilio:
Thanks!

PR #915, opened 2024-08-26
Added Godot Stammtisch Karlsruhe Website

nerdlibfront:
As we now have a website for our usergroup, i added that to the communities site. :)


coppolaemilio:
nice! :)

PR #917, opened 2024-08-28
Add new submissions page

coppolaemilio:
We realized that our submissions process for featuring new games on the showcase had been somewhat neglected. To address this, we’ve updated the criteria to make it clearer and more accessible, helping developers understand what’s expected before they submit their games.

Additionally, we’ve introduced a new submission form designed to make the process easier by asking targeted questions that help us filter and evaluate submissions.

## New submissions page

image
image

## Update on the current showcase page

image

### DIFF
--- a/pages/showcase/index.html
+++ b/pages/showcase/index.html
@@ -124,6 +124,8 @@ layout: default
 			</section>
 		</div>
 	</div>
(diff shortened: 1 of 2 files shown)


jams3223:
@coppolaemilio I vote for the white one with a dark grey header.

coppolaemilio:
@jams3223 They are both active! it just adapts to the user's theme. So if they prefer light pages, they will see the white one :)

PR #918, opened 2024-08-28
Fix Godot 4.4-dev1 download page preview

adamscott:
| Before | After |
| :---: | :---: |
| image | image |

@clayjohn, it seems that the ending slash in the URL is needed for previews to work properly.


coppolaemilio:
Thanks!

PR #922, opened 2024-08-31
Unify "Godot Engine" in html titles

JohnVeness:
Most pages have HTML title tags (which show in browser title bars, tabs, bookmarks etc,) with a suffix of "- Godot Engine", e.g.:

- Features - Godot Engine
- Contact - Godot Engine
- License - Godot Engine
- Download for Linux - Godot Engine

But some didn't conform, e.g.:

- Godot Engine - Blog
- Godot Engine - Community

or even just:

- Press Kit
- User Groups

This PR is an attempt to unify the page titles, including removing duplication of "Godot" when appropriate.

However, I'm not sure this is necessarily the best solution. Maybe the site name should be defined somewhere centrally and automatically added as a suffix, rather than added to pages manually as I have done.

Also, my solution doesn't fix pages like:

- https://godotengine.org/showcase/gourdlets/ which shows as "Gourdlets" not "Gourdlets - Godot Engine"
- https://godotengine.org/article/gamescom-2024/ which shows as "Gamescom here we come!" not "Gamescom here we come! - Godot Engine"
etc.

There may be other pages I've missed.

Also also, I'm not sure if the blog category pages should be like:
- "Blog - News - Godot Engine" as currently in the PR, or
- "News - Blog - Godot Engine", or even just
- "News - Godot Engine"

Hence marking this as a draft for now. Suggestions welcome!

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -23,8 +23,12 @@
 	<meta name="twitter:description" content="{% if page.description %}{{ page.description }}{% elsif page.excerpt %}{{ page.excerpt }}{% else %}Develop your 2D & 3D games, cross-platform projects, or even XR ideas!{% endif %}">
 	<meta name="twitter:image" content="{{ site.url }}{{ page.image | default: '/assets/share-image.webp' }}">
 
-	<title>{{ page.title | default: "Godot Engine" }}</title>
-
+	{% if page.title %}
+		<title>{{ page.title | remove: " - Godot Engine" }}{% if page.notitlesuffix != true %} – Godot Engine{% endif %}</title>
+	{% else %}
+		<title>Godot Engine</title>
+	{% endif %}
+	
 	<link rel="alternate" type="application/rss+xml" title="Godot News" href="/rss.xml">
 	<link rel="icon" href="/assets/favicon.png" sizes="any">
 	<link rel="icon" href="/assets/favicon.svg" type="image/svg+xml">
--- a/pages/community/community.html
+++ b/pages/community/community.html
@@ -1,6 +1,6 @@
 ---
 permalink: /community/index.html
-title: Godot Engine - Community
+title: Community
 layout: default
 ---
 
--- a/pages/education.html
+++ b/pages/education.html
@@ -1,4 +1,5 @@
 ---
+title: Education
 permalink: /education/index.html
 layout: default
 ---
--- a/pages/home.html
+++ b/pages/home.html
@@ -1,6 +1,7 @@
 ---
 permalink: /
 title: "Godot Engine - Free and open source 2D and 3D game engine"
+notitlesuffix: true
 description: "Godot provides a huge set of common tools, so you can just focus on making your game without reinventing
 the wheel."
 layout: default


coppolaemilio:
I like the idea of unifying the titles! instead of repeating them in every page this could be added on the template itself. So that when a title is provided, it gets added `+ ' – Godot Engine'`, Generally speaking, I think it is fine having or not having the "Godot Engine" on the title on every page, specially because as soon as you get a little happy with tabs, you can only partially see the first word, and the favicon is already a good indicator to know in which page you currently are.

JohnVeness:
> I like the idea of unifying the titles! instead of repeating them in every page this could be added on the template itself. So that when a title is provided, it gets added `+ ' – Godot Engine'`.

Thanks! If you could point me at the correct file to edit, I'll give it a go.

coppolaemilio:
Line 20 in the default.html contains the logic for the title tag in the <head>: https://github.com/godotengine/godot-website/blob/ce29cd143e54580be9389676672009f047a1263a/_layouts/default.html#L20

What you see there is the template checking if `page.title` exists to display it, or show the `Godot Engine` by default.
I imagine with something like:
```
{% if page.title %}
	<title>{{ page.title }} – Godot Engine</title>
{% else %}
	<title>Godot Engine</title>
{% endif %}
```

Could do the trick. But then you can be a bit more smart and also check if page.title already contains Godot Engine, or also add another variable to tell the template to not append the `– Godot Engine` part. 

Hope that helps!

JohnVeness:
Many thanks! I think I'm going to have to bail for now, though, as I'm having trouble getting the site to build locally so can't really test :(

Still, if you wanted to merge my manual changes anyway, at least that would be better, even if not perfect.

coppolaemilio:
@JohnVeness I might make a new PR doing this from the template side. But what issues were you running into? I've updated the readme to cover how to run this project in linux, but what OS are you using?

JohnVeness:
Ah, the new instructions seem to work fine, thanks. I expect the problem was that I wasn't using 3.1.2 - I was using 3.3, as supplied in the openSUSE Tumbleweed repos.

I'll have a fiddle with the template idea.

JohnVeness:
Please can you have a look now?

I have removed all the manual page title changes for now, and done everything in the template. This looks rather clumsy at the moment, because it first removes a suffix " - Godot Engine" if the page currently has it, then adds it on again! This is to avoid having to search and replace the rest of the site for now (I can do this in a separate PR or add it to this one, then simplify this).

I haven't done the same removal of a *prefix* "Godot Engine - " suffix yet. There are only a couple of pages that incorrectly use that, meaning you would see "Godot Engine - Community - Godot Engine" with this PR at the moment. Again, I can manually clean those in a separate PR or in this one.

I have added a "notitlesuffix" property that can be added to a page so as to not modify the title at all. I have added this property to the home page only, so that it still appears as "Godot Engine - Free and open source 2D and 3D game engine", as I assume that has been specifically chosen.

coppolaemilio (approved):
Updated a couple pages you missed. It should be good to go!

coppolaemilio:
Thanks!

JohnVeness:
Thanks for your help and encouragement! I quite enjoyed learning a little about jekyll.

I was wondering if similar should be done to the og:title and twitter:title meta tags? I don't particular wish to share a link onto that hellhole to see how they look though...

coppolaemilio:
maybe it is not necessary? still, you can get previews in some other platforms like discord or mastodon. If it works there, it is pretty much the same as in the rest of pages. Except linkedin... no one really knows what linkedin will do with your link 💩

PR #936, opened 2024-09-23
Download: Add Horizon OS APK and Horizon Store app, minor refactoring

akien-mga:
*Edit:* Removed that first change from the PR.
image

image

### DIFF
--- a/_data/download_platforms.yml
+++ b/_data/download_platforms.yml
@@ -14,7 +14,7 @@
 
 - name: "android.apk"
   title: "Android"
-  caption: "Universal APK"
+  caption: "APK - Universal"
   tags:
     - APK download
     - arm64
@@ -24,7 +24,7 @@
 
 - name: "android.playstore"
   title: "Android"
-  caption: "Play Store Universal"
+  caption: "Play Store"
   tags:
     - Play Store
     - arm64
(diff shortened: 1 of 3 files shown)


akien-mga:
Actually looking at my first screenshot, I made a mistake as it's linking the Horizon Store app on the `/download/android` page which is for 4.3-stable (but we have 4.4-dev2 on Horizon Store).

I'll remove it for now, and we should remember to add it when we release 4.4-stable.

akien-mga:
Added a fix for that stray dot too:
image

_data/download_platforms.yml
```diff
@@ -24,64 +24,74 @@
...
+    - Meta Quest 3 & Pro
+    - arm64
+
 - name: "linux.32"
   title: "Linux"
-  caption: "Standard (x86_32)"
+  caption: "x86_32"
   tags:
     - x86_32
-    - 32 bit
```
akien-mga, line 40: I removed the "32 bit" / "64 bit" tags, which don't say much in an age where most platforms are 64 bit and the relevant distinction is x86_64 or arm64.

_data/download_platforms.yml
```diff
@@ -24,64 +24,74 @@
...
+  title: "Horizon OS"
+  caption: "Horizon Store"
+  tags:
+    - Horizon Store
+    - Meta Quest 3 & Pro
+    - arm64
+
 - name: "linux.32"
   title: "Linux"
-  caption: "Standard (x86_32)"
```
akien-mga, line 37: I removed the "Standard" which didn't say anything useful. It's normally used to distinguish between "Standard" (no C#) and ".NET" (C#) builds, but the way this is designed, .NET builds were actually listed as "Standard - .NET" so all builds were "Standard".

coppolaemilio (approved):

coppolaemilio:
thanks!

PR #937, opened 2024-09-23
Download: Add link for Horizon Store app on stable release

akien-mga:
Depends on #936.

**To merge only when we release 4.4-stable.**

It adds a link to the `/download/android` page which is the one used for the current stable release. Currently it's 4.3 and the Horizon Store app is 4.4.dev2.

image

### DIFF
--- a/collections/_download/android.md
+++ b/collections/_download/android.md
@@ -13,6 +13,11 @@ downloads:
     featured: true
     featured_flavor: APK
 
+  - platform: "android.horizonstore"
+    custom: https://www.meta.com/en-gb/experiences/godot-game-engine/7713660705416473/
+    featured: true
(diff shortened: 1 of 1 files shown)


coppolaemilio:
What's the status on this one? should we salvage it?

akien-mga:
Rebased and tested (on top of #1023) and it seems to work fine.

Currently it shows our popup asking for donations before going to the Meta Store in the same tab, so the popup is only visible for a fraction of a second. We might want to tweak things so such external links get opened in a new tab maybe.

But that shouldn't be blocking for this.

coppolaemilio:
yeah, I can fix the popup in a followup PR

coppolaemilio (approved):
LGTM 🚀

PR #957, opened 2024-11-17
Remove PLC reference from the contact page

adamscott:
As there's both the governance page AND a dedicated website (https://godot.foundation), I don't think we need to update this part for the Board.


coppolaemilio:
thanks!

PR #963, opened 2024-11-21
Style: Modernize yaml formatting & syntax

Repiteo:
Applies the same formatting fixes used by the main repo. Also bumped the checkout action version because v3 was throwing warnings


coppolaemilio (changes requested):
Hey, if the issue is on the action you should make a PR only for fixing that. Jekyll is extremely finicky and I don't think there is any value in changing all the .yml files of the project since they work fine.

coppolaemilio:
Closing

PR #982, opened 2024-12-20
Add full content on Atom feed

martgnz:
Atom does have a different tag for content and summary, so let's use it to render the full posts.


coppolaemilio:
I'll merge this one as-is, don't have a reader to confirm that all the posts look properly

PR #993, opened 2025-01-14
Add blog post about UID changes coming to Godot 4.4

Calinou:
Cover image source: [uid-changes-coming-to-godot-4-4.xcf.zip](https://github.com/user-attachments/files/18414040/uid-changes-coming-to-godot-4-4.xcf.zip)

<details>
<summary>Text used in cover image</summary>

```gdscript
extends Node

const A = preload("uid://cevl8wq8efujh")
const B = preload("uid://bnaqmivrjyrm1")

var c = load("uid://b3womovlq200m")
var d = load("uid://q5x2bsae1kci")
```
</details>


collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+excerpt: "Godot's resource UID system is changing to better support scripts and shaders. Read on to know what's required on your part to fully benefit from it."
+categories: ["progress-report"]
+author: Hugo Locurcio
+image: /storage/blog/covers/uid-changes-coming-to-godot-4-4.webp
+date: 2025-01-15 14:00:00
+---
+
+## Why use UIDs in Godot?
+
+Until Godot 4.0, the engine exclusively relied on file paths to refer to scenes and scripts. While this means references in scenes and scripts use human-readable paths, this means that these paths will break if files are moved to a different location in the project (regardless of whether it was done in the FileSystem dock or outside Godot). This is especially problematic for large projects where filesystem organization often changes regularly.
```
JohnVeness:
```suggestion
Until Godot 4.0, the engine exclusively relied on file paths to refer to scenes and scripts. While this means references in scenes and scripts use human-readable paths, these paths will break if files are moved to a different location in the project (regardless of whether the move was done using the FileSystem dock or outside Godot). This is especially problematic for large projects where filesystem organization often changes regularly.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+
+Since Godot 3.x, the editor already updates path references if files are moved within the FileSystem dock. This works more or less, but has drawbacks: The process is slow in large projects (as files need to be scanned and modified), and a single change results in large amount of modified files in commits. Additionally, it is common for users to move these files outside Godot (e.g. using the OS file manager or the command line).
```
JohnVeness:
```suggestion
Since Godot 3.x, the editor already updates path references if files are moved within the FileSystem dock. This works more or less, but has drawbacks: the process is slow in large projects (as files need to be scanned and modified), and a single change results in large amount of modified files in commits.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+
+If the editor is not open, it has no way to know the change happened. Even if the editor is open, when working with version control, files may change or be moved and the editor has no way to know where files have gone, resulting in all sorts of errors.
```
JohnVeness:
```suggestion
Additionally, it is common for users to move these files outside Godot (e.g. using the OS file manager or the command line). If the editor is not open, it has no way to know the change happened. Even if the editor is open, when working with version control, files may change or be moved and the editor has no way to know where files have gone, resulting in all sorts of errors.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Limitations of UIDs before Godot 4.4
+
+Each imported resource in Godot has a `.import` file, which allows for storing an UID. Additionally, Godot's own scene and resource file formats store an UID in their header. However, scripts and shaders are not imported resources and they are "plain text" formats, without any Godot-specific information stored within. In practice, this meant that UIDs could be used for scenes and most resources, but not scripts and shaders. The UID implementation was therefore not covering every resorce type Godot support ([scripts](https://docs.godotengine.org/en/stable/classes/class_script.html) and [shaders](https://docs.godotengine.org/en/stable/classes/class_shader.html) are resources too).
```
JohnVeness:
```suggestion
Each imported resource in Godot has a `.import` file, which allows for storing an UID. Additionally, Godot's own scene and resource file formats store an UID in their header. However, scripts and shaders are not imported resources and they are "plain text" formats, without any Godot-specific information stored within. In practice, this meant that UIDs could be used for scenes and most resources, but not scripts and shaders. The UID implementation was therefore not covering every resource type Godot support ([scripts](https://docs.godotengine.org/en/stable/classes/class_script.html) and [shaders](https://docs.godotengine.org/en/stable/classes/class_shader.html) are resources too).
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+
+Thanks to [GH-100803](https://github.com/godotengine/godot/pull/100803), you can see the underlying `res://` path and resource type when hovering the `uid://` path with the mouse in the script editor. (This feature also works when hovering `res://` paths, so it is not exclusive to `uid://` paths.)
+
+## What should I change in my project or workflow?
+
+The most important step is to **make sure `.uid` files are committed to version control**. In other words, `*.uid` should **not** be added to `.gitignore`.
+
+Secondly, when you move a script or shader outside Godot, **you should also move the `.uid` file along with it**. When you remove a script or shader outside Godot, you should also remove the `.uid` file, although this is less important – not removing the `.uid` file won't break anything.
+
+Moving files externally will produce warnings when opening scenes referring said files. You need to re-save all depending scenes to remove warnings.
```
JohnVeness:
```suggestion
Moving files externally will produce warnings when opening scenes referring to said files. You need to re-save all depending scenes to remove warnings.
```
or
```suggestion
Moving files externally will produce warnings when opening scenes referencing said files. You need to re-save all depending scenes to remove warnings.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,87 @@
...
+
+While using separate UID files is not perfect by any means, we felt it was a better compromise for Godot's design and to ensure version control remains as seamless as possible.
+
+### What happens if I don't commit `.uid` files to version control?
+
+If you forget to commit `.uid` files to version control, the project will still work locally. However, as soon as you clone the project on another device, the UID references will break. Godot will still be able to resolve them if the paths haven't changed (since paths are also saved as a fallback), but a warning will be printed for every resource where a path fallback had to be used to infer the reference. This behavior should not be relied upon intentionally, as it is possible for two resources to be swapped around without any of their paths changing. This can occur if one of the resources' paths was changed to be identical to another resource that was relocated elsewhere.
```
JohnVeness:
```suggestion
If you forget to commit `.uid` files to version control, the project will still work locally. However, as soon as you clone the project on another device, the UID references will break. Godot will still be able to resolve them if the paths haven't changed (since paths are also saved as a fallback), but a warning will be printed for every resource where a path fallback had to be used to infer the reference. This behavior should not be relied upon intentionally, as it is possible for two resources to be swapped around without either of their paths changing. This can occur if the path of one of the resources was changed to be identical to another resource that was relocated elsewhere.
```

coppolaemilio:
Amazing, thank you so much @Calinou !
I think we should add to the FAQ the "Why not making a `.meta` file instead?" Which is something that comes up again and again.

Calinou:
> think we should add to the FAQ the "Why not making a `.meta` file instead?" Which is something that comes up again and again.

I added this to the FAQ.

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+excerpt: "Godot's resource UID system is changing to better support scripts and shaders. Read on to know what's required on your part to fully benefit from it."
+categories: ["progress-report"]
+author: Hugo Locurcio
+image: /storage/blog/covers/uid-changes-coming-to-godot-4-4.webp
+date: 2025-01-15 14:00:00
+---
+
+## Why use UIDs in Godot?
+
+Until Godot 4.0, the engine exclusively relied on file paths to refer to scenes and scripts. While this means references in scenes and scripts use human-readable paths, these paths will break if files are moved to a different location in the project (regardless of whether the move was done using the FileSystem dock or outside Godot). This is especially problematic for large projects where filesystem organization often changes regularly.
```
akien-mga:
```suggestion
Until Godot 4.0, the engine exclusively relied on file paths to refer to scenes, scripts and other resources. While this means references in scenes and scripts use human-readable paths, these paths will break if files are moved to a different location in the project (regardless of whether the move was done using the FileSystem dock or outside Godot). This is especially problematic for large projects where filesystem organization often changes regularly.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+Since Godot 3.x, the editor already updates path references if files are moved within the FileSystem dock. This works more or less, but has drawbacks: the process is slow in large projects (as files need to be scanned and modified), and a single change results in large amount of modified files in commits.
```
akien-mga:
```suggestion
Since Godot 3.x, the editor already updates path references if files are moved within the FileSystem dock. This works more or less, but has drawbacks: the process is slow in large projects (as files need to be scanned and modified), file paths in scripts are left unchanged, and a single file move can result in a large amount of modified files in commits. It's also not fully reliable, and veteran Godot users probably remember various situations where moving a file seemingly corrupted a scene or resource.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+Additionally, it is common for users to move these files outside Godot (e.g. using the OS file manager or the command line). If the editor is not open, it has no way to know the change happened. Even if the editor is open, when working with version control, files may change or be moved and the editor has no way to know where files have gone, resulting in all sorts of errors.
```
akien-mga:
```suggestion
Additionally, it is common for users to move these files outside Godot (e.g. using the OS file manager, an IDE or the command line). If the editor is not open, it has no way to know the change happened. Even if the editor is open, when working with version control, files may change or be moved and the editor has no way to know where files have gone, resulting in all sorts of errors.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+With the aim of making Godot better suitable to large projects and significantly more resilient to external filesystem changes, ensuring this is a fully supported workflow has become a priority. External file manager programs often have more features than Godot's own FileSystem dock, making them more convenient for many use cases. This means we need an approach that can reliably track files being moved, even if the Godot editor isn't open.
```
akien-mga:
```suggestion
With the aim of making Godot better suited for large projects and significantly more resilient to external filesystem changes, ensuring this is a fully supported workflow has become a priority. External file manager programs often have more features than Godot's own FileSystem dock, making them more convenient for many use cases. This means we need an approach that can reliably track files being moved, even if the Godot editor isn't open.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Limitations of UIDs before Godot 4.4
+
+Each imported resource in Godot has a `.import` file, which allows for storing an UID. Additionally, Godot's own scene and resource file formats store an UID in their header. However, scripts and shaders are not imported resources and they are "plain text" formats, without any Godot-specific information stored within. In practice, this meant that UIDs could be used for scenes and most resources, but not scripts and shaders. The UID implementation was therefore not covering every resource type Godot support ([scripts](https://docs.godotengine.org/en/stable/classes/class_script.html) and [shaders](https://docs.godotengine.org/en/stable/classes/class_shader.html) are resources too).
```
akien-mga: Since it's an initialism it's "a you-eye-dee".
```suggestion
Each imported resource in Godot has a `.import` file, which allows for storing a UID. Additionally, Godot's own scene and resource file formats store a UID in their header. However, scripts and shaders are not imported resources and they are "plain text" formats, without any Godot-specific information stored within. In practice, this meant that UIDs could be used for scenes and most resources, but not scripts and shaders. The UID implementation was therefore not covering every resource type Godot support ([scripts](https://docs.godotengine.org/en/stable/classes/class_script.html) and [shaders](https://docs.godotengine.org/en/stable/classes/class_shader.html) are resources too).

Moreover, the editor support for making use of UIDs and relying primarily on them was lacking, so UIDs didn't yet fulfill their promise of enabling seamless refactoring of project files.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+Thanks to [GH-100803](https://github.com/godotengine/godot/pull/100803), you can see the underlying `res://` path and resource type when hovering the `uid://` path with the mouse in the script editor. (This feature also works when hovering `res://` paths, so it is not exclusive to `uid://` paths.)
+
+## What should I change in my project or workflow?
+
+The most important step is to **make sure `.uid` files are committed to version control**. In other words, `*.uid` should **not** be added to `.gitignore`.
+
+Secondly, when you move a script or shader outside Godot, **you should also move the `.uid` file along with it**. When you remove a script or shader outside Godot, you should also remove the `.uid` file, although this is less important – not removing the `.uid` file won't break anything.
+
+Moving files externally will produce warnings when opening scenes referencing said files. You need to re-save all depending scenes to remove warnings.
```
akien-mga, line 56: I'd argue this is probably a bug that we should fix. If one needs to open all scenes and re-save them after moving a file, we haven't improved the workflow much. Or do I misunderstand? CC @KoBeWi

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Why not use replace `.import` files with `.meta` files and use them for all resources that have an UID?
+
+While Godot must import many resource types before they can be used in a project, not all types actually need to be imported to be used. Resources that don't need to be imported to be used don't have properties that you could set on import, as they are directly loaded by Godot with no remapping system in place. Instead, everything about the file is inferred from the file itself. This approach makes more sense for files that you expect to be able to copy-paste around (or even copy directly from websites), while making sure the file remains functional without needing a separate metadata file. It's a common workflow from users to copy scripts or shaders from websites such as [Godot Shaders](https://godotshaders.com/) after all.
+
+The main examples of resources that don't need to be imported are scripts and shaders, hence the dedicated `.uid` files.
```
akien-mga, line 74:
```suggestion
The main examples of resources that don't need to be imported are scripts and shaders, hence the dedicated `.uid` files.

It has been suggested that we could still use `.meta` files for scripts and shaders, and use them to store more information than just UIDs, such as user-defined metadata. For the time being we prefer to focus on the problem at hand (the need of UIDs for better refactoring possibility) and not future proof things. If we ever change our mind, it would be easy enough to migrate information from `.uid` files to whichever new container we decide to use.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+  - In contrast, using separate files ensures merge conflicts don't happen quite as often. If conflicts do arise, they can be resolved by just including the file (or removing it if the source file has also been removed).
+- Centralizing UID information is problematic for add-ons. For example, if a developer refactors their addon and moves files around, then someone downloads the new version of their addon into an existing project of theirs, references to files in the addon in their project get updated automatically. If UIDs are stored in a centralized fashion, there wouldn't be a way to do this as the addon's UIDs wouldn't be stored in the addon. This would result in UID references breaking.
+- If you have multiple Godot editor instances open on the same project, each instance may write a different version of the file, leading to conflicts that can only be detected once you reopen the editor. Additionally, on Windows, you would get a lot of errors while using the editor due to file locking limitations.
+
+While using separate UID files is not perfect by any means, we felt it was a better compromise for Godot's design and to ensure version control remains as seamless as possible.
```
akien-mga: This is missing the main reason IMO: with a centralized database, you can't handle moving files externally to Godot.

If you move a script or shader outside Godot, the reference to it in the centralized mapping would now point at a missing file path, and it wouldn't know the new file path. That's why the UID information must be in or close to the file, so it's moved alongside it.

I'd suggest adding this as the first point, and maybe removing the last point which is a bit more anecdotal (and could be worked around in some way with process-specific temporary files that get written to the main file on exit for example).

akien-mga (commented):
Looks great overall.

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+
+### What happens if I don't commit `.uid` files to version control?
```
akien-mga, line 99:
```suggestion
### Why not embed the UID directly in the scripts as a comment or annotation?

We evaluated it, but we saw a lot of pushback against having this kind of magic string injected as a comment or annotation in all script files, as testers found it to be quite disruptive.

On the technical side, there are also implementation hurdles. All scripting languages supported by Godot (including community bindings via GDExtension) would need to implement this kind of magic comment or annotation (if they have such concept) with an appropriate syntax. The Godot editor would need to know how to extract these UIDs from all types of scripts in an efficient manner (without having to parse it).

Additionally, external IDEs and tools may not play nice with such Godot-specific comments or annotations in various scripting languages, and if Godot-specific plugins want to implement UID support, they would also need to manually implement how to extract the UIDs from the scripts in an efficient manner (as opposed to just reading a file with a predictable name).

Finally, it's not only scripts and shaders that need UIDs, but all resources. While most other builtin resource types can store UID information as described above, this system is also relevant for custom resource formats that users can implement. Having `.uid` files generated for such resources makes the system work out of the box.

### What happens if I don't commit `.uid` files to version control?
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -0,0 +1,93 @@
...
+  - In contrast, using separate files ensures merge conflicts don't happen quite as often. If conflicts do arise, they can be resolved by just including the file (or removing it if the source file has also been removed).
+- Centralizing UID information is problematic for add-ons. For example, if a developer refactors their addon and moves files around, then someone downloads the new version of their addon into an existing project of theirs, references to files in the addon in their project get updated automatically. If UIDs are stored in a centralized fashion, there wouldn't be a way to do this as the addon's UIDs wouldn't be stored in the addon. This would result in UID references breaking.
+- If you have multiple Godot editor instances open on the same project, each instance may write a different version of the file, leading to conflicts that can only be detected once you reopen the editor. Additionally, on Windows, you would get a lot of errors while using the editor due to file locking limitations.
```
akien-mga: As suggested here: https://github.com/godotengine/godot-website/pull/993/files#r1915694264

The diff is messy, I'm just adding a new first bullet point, and removing the last one, which I think is a bit too anecdotal, and this avoids this FAQ item from being way to big.
```suggestion
- With a centralized file for all UIDs, the purpose of being able to move scripts or shaders outside of Godot would not be fulfilled. When moving a script outside of Godot, the centralized UID mapping would not be updated, and thus it would lose the connection between the previous path and its UID. For this use case to be supported, the UID needs to be close to the file being moved - either alongside it in a `.uid` or `.import` file, or embedded in it (e.g. for `.tscn` or `.tres` files).
- This file would be subject to frequent merge conflicts in version control systems. For example, if you have two branches that add or edit different scripts/shaders, they would each add their own lines to the UID database file. Both of these branches would add these lines at the same location in the file (probably at the end). However, once you try to merge those two branches together, your version control system wouldn't be able to know what to keep automatically. This is where a merge conflict occurs.
  - In contrast, using separate files ensures merge conflicts don't happen quite as often. If conflicts do arise, they can be resolved by just including the file (or removing it if the source file has also been removed).
- Centralizing UID information is problematic for add-ons. For example, if a developer refactors their addon and moves files around, then someone downloads the new version of their addon into an existing project of theirs, references to files in the addon in their project get updated automatically. If UIDs are stored in a centralized fashion, there wouldn't be a way to do this as the addon's UIDs wouldn't be stored in the addon. This would result in UID references breaking.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
akien-mga: Added in https://github.com/godotengine/godot-website/pull/993#pullrequestreview-2551182572.

akien-mga (approved):

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -71,17 +73,29 @@ While Godot must import many resource types before they can be used in a project
 
 The main examples of resources that don't need to be imported are scripts and shaders, hence the dedicated `.uid` files.
 
+It has been suggested that we could still use `.meta` files for scripts and shaders, and use them to store more information than just UIDs, such as user-defined metadata. For the time being we prefer to focus on the problem at hand (the need of UIDs for better refactoring possibility) and not future proof things. If we ever change our mind, it would be easy enough to migrate information from `.uid` files to whichever new container we decide to use.
```
JohnVeness:
```suggestion
It has been suggested that we could still use `.meta` files for scripts and shaders, and use them to store more information than just UIDs, such as user-defined metadata. For the time being we prefer to focus on the problem at hand (the need of UIDs for better refactoring possibility) and not future-proof things. If we ever change our mind, it would be easy enough to migrate information from `.uid` files to whichever new container we decide to use.
```

collections/_article/uid-changes-coming-to-godot-4-4.md
```diff
@@ -71,17 +73,29 @@ While Godot must import many resource types before they can be used in a project
...
-- If you have multiple Godot editor instances open on the same project, each instance may write a different version of the file, leading to conflicts that can only be detected once you reopen the editor. Additionally, on Windows, you would get a lot of errors while using the editor due to file locking limitations.
 
 While using separate UID files is not perfect by any means, we felt it was a better compromise for Godot's design and to ensure version control remains as seamless as possible.
 
+### Why not embed the UID directly in the scripts as a comment or annotation?
+
+We evaluated it, but we saw a lot of pushback against having this kind of magic string injected as a comment or annotation in all script files, as testers found it to be quite disruptive.
+
+On the technical side, there are also implementation hurdles. All scripting languages supported by Godot (including community bindings via GDExtension) would need to implement this kind of magic comment or annotation (if they have such concept) with an appropriate syntax. The Godot editor would need to know how to extract these UIDs from all types of scripts in an efficient manner (without having to parse it).
```
JohnVeness:
```suggestion
On the technical side, there are also implementation hurdles. All scripting languages supported by Godot (including community bindings via GDExtension) would need to implement this kind of magic comment or annotation (if they have such a concept) with an appropriate syntax. The Godot editor would need to know how to extract these UIDs from all types of scripts in an efficient manner (without having to parse it).
```

akien-mga:
Thanks!

PR #999, opened 2025-01-18
Fix incorrect link on preview downloads page

JimmyCushnie:
Edit: #1000 is an alternative that's probably better, since `#stable` is a better anchor for this section of this page than `#call-to-action`.

---

There's a link at the top of the [previews download page](https://godotengine.org/download/preview/) which says "Looking for a stable version? See below!"

This link is supposed to scroll the page to the section lower down where you can download the stable version. However, currently this link doesn't work because it points to an anchor that didn't exist.

The link points to `#stable`. However, the actual anchor, [as defined](https://github.com/godotengine/godot-website/blob/5d28982dffd98030300ec87582f1d352453e78c6/_includes/download/download-section.html#L60C10-L60C24) in `download-section.html`, is `#call-to-action`.

This PR fixes the link so that clicking it works properly.

### DIFF
--- a/pages/download/preview.html
+++ b/pages/download/preview.html
@@ -41,7 +41,7 @@ layout: default
 
 	<div class="other-platforms">
 		<div class="other-platforms-wrapper">
-			Looking for the stable version? <a href="#stable">See below!</a>
+			Looking for the stable version? <a href="#call-to-action">See below!</a>
 		</div>
 	</div>
 </div>


JohnVeness:
I wonder if it would be better if the anchor was changed? https://godotengine.org/download/preview/#stable seems a better URL than https://godotengine.org/download/preview/#call-to-action

JimmyCushnie:
I agree actually. But since the download widget is reused in several places on the website, I didn't want to break other links to its anchor (our internal links or any external links that might be out there).

If we're going to change the download widget's anchor, my vote is `#download-latest-stable`. But that would be a bigger PR and would require careful checking for other links that need to be updated.

JohnVeness:
Yeah, I agree we would have to tread carefully.

Maybe `{% include /download/download-section.html title="Need a stable version?" %}` can be changed to something like `{% include /download/download-section.html title="Need a stable version?" anchor="stable" %}` and download-section.html use include.anchor in the appropriate place, only using `#call-to-action` as a fallback if the anchor wasn't specified.

Regarding `#stable` vs `#download-latest-stable`, presumably https://godotengine.org/download/preview/#stable used to work prior to commit bb0f0a4, so it would be nice to go back to that in order not to break any old external links or bookmarks.

JimmyCushnie:
> Maybe `{% include /download/download-section.html title="Need a stable version?" %}` can be changed to something like `{% include /download/download-section.html title="Need a stable version?" anchor="stable" %}` and download-section.html use include.anchor in the appropriate place, only using `#call-to-action` as a fallback if the anchor wasn't specified.

Nice, this sounds like a great solution to me. Thanks for the historical context as well.

May I go ahead and implement it?

JohnVeness:
> May I go ahead and implement it?

I'm not a decision maker round here, but I expect you could!

JimmyCushnie:
Cheers. It took me a bit to get the site building locally for testing but I just opened #1000 with your suggested solution.

JimmyCushnie:
Superseded by #1000

PR #1000, opened 2025-01-20
Add custom anchor to download-section.html, and use it to fix link on previews download page

JimmyCushnie:
Alternative to #999.

There's a link at the top of the [previews download page](https://godotengine.org/download/preview/) which says "Looking for a stable version? See below!"

This link is supposed to scroll the page to the section lower down where you can download the stable version. However, currently this link doesn't work because it points to an anchor that no longer exists.

The old anchor, `#stable`, was [removed](https://github.com/godotengine/godot-website/commit/bb0f0a41c3cb1a69dbf7f96ce9c4d98eb9e55822#diff-48f02e6e05568e0d339b6cce00e810761d7046d12d88c855796f6c19ac6269b9L127) in https://github.com/godotengine/godot-website/commit/bb0f0a41c3cb1a69dbf7f96ce9c4d98eb9e55822, when `download-section.html` was consolidated.

This PR lets you set a custom anchor when you include `download-section.html`, and does so in `preview.html` to restore the broken link.

### DIFF
--- a/_includes/download/download-section.html
+++ b/_includes/download/download-section.html
@@ -57,7 +57,7 @@
 		color: white;
 	}
 </style>
-<div id="call-to-action" class="container card section-download">
+<div id="{{ include.anchor | default: 'call-to-action' }}" class="container card section-download">
 	{% if include.title %}
 		<h2>{{ include.title }}</h2>
 	{% else %}
--- a/_includes/download/download-section.html
+++ b/_includes/download/download-section.html
@@ -58,11 +58,7 @@
 	}
 </style>
 <div id="{{ include.anchor | default: 'call-to-action' }}" class="container card section-download">
-	{% if include.title %}
-		<h2>{{ include.title }}</h2>
-	{% else %}
-		<h2>Ready to start?</h2>
-	{% endif %}
+    <h2>{{ include.title | default: 'Ready to start?' }}</h2>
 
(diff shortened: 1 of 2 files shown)


JohnVeness:
I have tested locally and this works great. The only other page that includes download-section.html is features.html, and that still works fine (still uses `#call-to-action`, although as far as I can tell nothing internal links to that anyway).

@coppolaemilio Do you want to check this?

_includes/download/download-section.html
```diff
@@ -57,7 +57,14 @@
 		color: white;
 	}
 </style>
-<div id="call-to-action" class="container card section-download">
+<div 
+    {% if include.anchor %}
+        id="{{ include.anchor }}"
```
coppolaemilio: I believe there was a way to set up a default value if a variable is not found. I can look it up later on, or if you are able to find it it will simplify this into just one line.

JimmyCushnie: Thanks very much! I used this feature for the anchor as you suggested :) I also noticed it could be used for the header right nearby, so I went ahead and did it there as well.

coppolaemilio (approved):

coppolaemilio:
Awesome! thanks :)

JimmyCushnie:
Woohoo! Happy PR #1000 everyone!

PR #1003, opened 2025-01-28
Fix Editor section formatting in `2024 Cherry-Picks` blogpost

syntaxerror247:
Keeping `Android Editor` in next line would look better. Also added link to the PR that added export support.


coppolaemilio:
I feel like this change would break the style of the list. Maybe adding a `, and` would help to make it stand out more instead?

syntaxerror247:
Okay, I have replaced `<br>` with `and`.

PR #777, opened 2024-01-01
Sync Maintenance links to Community page

fellowjamess:
Removed Q&A link and changed it to the new forum link.

PR #555, opened 2023-02-11
Make the main scrollbar follow user's color preference

Ymanawat:
Resolved the issue of the scrollbar having only light theme.

- *Bugsquad edit: This closes https://github.com/godotengine/godot-website/issues/547.*

PR #1008, opened 2025-02-07
Improve footer text readability

enkvadrat:


coppolaemilio:
Can you show before and after screenshots?

enkvadrat:
Before:
image
After:
image

Calinou (approved):
The light font weight was always difficult to read me for in the footer. Thanks for fixing it :slightly_smiling_face: 

Note that depending on your OS/browser, it's possible the font never appeared as light in the first place (if the browser tried to use a font without a light weight). It did on my Fedora 41 + Firefox setup though.

coppolaemilio:
@enkvadrat does 300 instead of 400 look good on your end? I find 400 to be too bold in my displays, but 300 is nicer

Calinou:
> @enkvadrat does 300 instead of 400 look good on your end? I find 400 to be too bold in my displays, but 300 is nicer

This is what it looks like on my end on Firefox on Fedora 41:

### `font-weight: 200`

Light | Dark
-|-
image | image

### `font-weight: 300`

Light | Dark
-|-
image | image

### `font-weight: 400`

Light | Dark
-|-
image | image

`font-weight: 300` is acceptable, but I think `font-weight: 400` is significantly more accessible to users with low vision. I doubt `font-weight: 300` would pass WCAG 2.0 standards here. As a bonus, `font-weight: 400` fixes the awkward line break in the footer, although we can (and should) do that manually too.

I think ideally, we'd use `font-weight: 400` but change the text color to make it slightly more faded. This way, the text is easy enough to read but doesn't take too much attention away from the rest of the page:

Light (`#555`) | Dark (`#777`)
-|-
image | image

The color is applied on `footer.footer-global a`.

coppolaemilio:
Thank you @Calinou for the screenshots! I think I'll merge this and then make a future commit to fix it on Mac displays, because the weights do look different in my screen:
The current one
image

300
image

400
image

coppolaemilio:
Thank you @enkvadrat for your first contribution! :)

PR #1022, opened 2025-02-28
Change the display name of `zh-cn` and `zh-tw` locale

timothyqiu:
Rationale: https://github.com/godotengine/godot-proposals/issues/6289

In short, the current wording could cause potential problems. This situation is more awkward than the wording in the editor, because many non-Godot users will see it.

This PR uses the same strategy as Microsoft, distinguishing by script rather than region.

| Before | After |
|---|---|
| Chinese (China) | Chinese (Simplified) |
| Chinese (Taiwan) | Chinese (Traditional) |

### DIFF
--- a/assets/js/localize.js
+++ b/assets/js/localize.js
@@ -5,8 +5,8 @@ const languageMap = {
 	'de': 'Deutsch',
 	'fr': 'Français',
 	'ja': '日本語',
-	'zh-cn': '中文（中国）',
-	'zh-tw': '中文（台灣）',
+	'zh-cn': '中文（简体）',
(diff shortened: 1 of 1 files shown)


coppolaemilio:
thanks! sounds good, I'm not so well versed into this topic, so appreciate the input

SheepYhangCN:
What about change ```cn``` and ```tw``` to ```hans``` and ```hant```?
I think consistency is more important

timothyqiu:
I think both approaches have their pros and cons. The main reason I did not change the locale identifier is that the corresponding page is already live. Changing the locale would change the URL of the page, causing unnecessary inconvenience to users.

PR #1027, opened 2025-03-03
Add 4.4 banner to the main page.

adamscott:
Add 4.4 banner to the main page. This will be easily reusable for other releases or events.

image

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,5 +1,5 @@
 <input type="checkbox" id="nav_toggle_cb">
-<header>
+<header class="flex column">
 	<div class="container flex align-center">
 		<div id="nav_head">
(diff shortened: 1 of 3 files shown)


Calinou:
The banner looks good, but unfortunately, it breaks rendering on mobile. The right of the page has nothing inside and the "hamburger" menu icon is cut off.

iPhone SE 2nd Gen on Firefox:

Before | After
-|-
image | image

iPhone 12 Pro on Chromium:

Before | After
-|-
image | image

Additionally, I suggest using smaller text on mobile so that it's more likely to fit on a single line.

adamscott:
@Calinou Fixed!

https://github.com/user-attachments/assets/8deef9ee-fe28-416b-b71c-f469d5d21d29

PR #1030, opened 2025-03-04
Add pre-commit and codespell check

adamscott:
This PR adds codespell and pre-commit to the repo.
(it also adds .venv for local Python environments to .gitignore)


coppolaemilio:
This can be revisited when we split the blog repository from the main website. Thanks!

PR #1041, opened 2025-03-21
Fix download card donate link color on light theme

Calinou:
Previously, it appeared as dark gray text on a blue background, which was difficult to read. It now appears as white text on a blue background.

The appearance on dark theme is unchanged.

## Preview

Before | After
-|-
image | image

___

Light theme | Dark theme (unchanged)
-|-
image | image


coppolaemilio:
Thanks!

PR #1049, opened 2025-03-28
Add Godot User Group Berlin website

mtttz:
The Godot User Group Berlin has a new website: https://godot.berlin


coppolaemilio:
nice!

PR #1058, opened 2025-05-07
Add Godot Seattle to communities.yml

TranquilMarmot:

PR #1061, opened 2025-05-08
Add note to user groups page with info on where to open a PR to add a group

TranquilMarmot:
Over in the Godot Seattle group, it sounds like somebody emailed `webmaster@godotengine.org` but we never got a response. It seems like adding directly to the `_data/communities.yml` is the more "correct" way to do things (see also https://github.com/godotengine/godot-website/pull/1058 )

image


coppolaemilio (approved):

coppolaemilio:
Thanks!

PR #1065, opened 2025-05-14
Add automatic browser language selector

adamscott:
This PR adds an automatic redirection to the right language considering the browser's preferred languages. See the first comment for the algorithm.

https://github.com/user-attachments/assets/70de1fcc-d591-4ce8-a12d-7658f9c8b773

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -32,7 +32,7 @@
 						<a class="mobile-language-selector" href="#">{% t header.language %}: </a>
 						<div class="language-dropdown">
 							{% for lang in page.localize %}
-								<div class="language-option"><a href="#" data-lang-path="{% if lang == 'en' %}/{% else %}/{{ lang }}/{% endif %}" data-lang="{{ lang }}" onclick="setLanguagePreference(event, this)"><span class="localize-language-label">{{ lang }}</span></a></div>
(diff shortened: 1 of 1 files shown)


_includes/header.html
```diff
@@ -60,32 +60,142 @@
...
+			const browserPreferredLanguage = getBrowserPreferredLanguage();
+			if (browserPreferredLanguage != null) {
+				setPreferredLanguage(browserPreferredLanguage);
+				redirectTo(getLanguagePath(browserPreferredLanguage));
+			}
+		} else if (preferredLanguage !== currentLanguage && currentLanguage === "en") {
+			redirectTo(getLanguagePath(preferredLanguage));
+		} else {
+			setPreferredLanguage(currentLanguage);
+		}
```
adamscott: Algorithm for language selection (on page load).

_includes/header.html
```diff
@@ -60,32 +60,154 @@
...
+		return _languageSelector;
+	}
+
+	function setPreferredLanguage(lang) {
+		window.localStorage.setItem("preferred_language", lang);
+	}
+
+	function getPreferredLanguage() {
+		return window.localStorage.getItem("preferred_language");
+	}
```
winston-yallow, line 103: Do you think we can use cookies here? We may want to redirect server side on initial page load, so it would be nice if we can access the preferred language there too.

winston-yallow (changes requested):

coppolaemilio:
Thanks! In the future, it would be nice to do this server side and change to using cookies.

PR #1063, opened 2025-05-14
Add "Upcoming (serious) Web performance boost" article

adamscott:
image

PR #1077, opened 2025-06-05
Corrected RAM stats typo

aidan-j-rhoden:
I know this is a minuscule nitpick, but the RAM rating `C30` should be `CL30`, as it stands for CAS-Latency (Column Address Strobe)

Thanks as always for updates and such performance improvements, I'm looking forward to trying it out!

- *Production edit: Follow-up to https://github.com/godotengine/godot-website/pull/1063.*

### DIFF
--- a/collections/_article/upcoming-serious-web-performance-boost.md
+++ b/collections/_article/upcoming-serious-web-performance-boost.md
@@ -30,7 +30,7 @@ Our resident benchmark expert Hugo Locurcio (better known as [Calinou](https://g
 
 - **CPU:** Intel Core i9-13900K
 - **GPU:** NVIDIA GeForce RTX 4090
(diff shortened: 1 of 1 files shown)


Calinou (approved):
While `CAS` or even `C` are often used as a generic term for CAS latency, it's more correct to say CL. Actual RAM latency is a factor of several timing values (not just CL) as well as its frequency.

aidan-j-rhoden:
> While `CAS` or even `C` are often used as a generic term for CAS latency, it's more correct to say CL. Actual RAM latency is a factor of several timing values (not just CL) as well as its frequency.

Thanks!  I did double-check to see if `C` was used as well, but it didn't seem very common.

coppolaemilio:
Thanks!

PR #1079, opened 2025-06-12
Fix building with Jekyll on Ruby 3.4

Calinou:
Ruby 3.4 removed several standard library packages, so they have to be installed as gems instead.

This also updates Jekyll to use a version that's supported in Ruby 3.4.

The minimum required Ruby version is now 3.2.

Many deprecation warnings occur when building Sass, which I've fixed in a separate PR that depends on this one: https://github.com/godotengine/godot-website/pull/1080

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -18,10 +18,10 @@ jobs:
 
     # Configure the build environment.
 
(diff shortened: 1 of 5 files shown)


adamscott (approved):

coppolaemilio:
Thanks!

PR #1089, opened 2025-06-23
Add `logger` to Gemfile to fix Ruby 3.5.0 deprecation warning

Calinou:
- Follow-up to https://github.com/godotengine/godot-website/pull/1079.

[The `logger` gem](https://rubygems.org/gems/logger) will be moved out of the standard library in Ruby 3.5.0. While the website can still build fine with Ruby 3.4.x, this won't be the case anymore with the latest Ruby version once 3.5.0 releases.

Previously, this was printed when running `bundle exec jekyll s`:

```
/home/hugo/.local/share/gem/ruby/3.4.0/gems/jekyll-4.4.1/lib/jekyll.rb:26: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,3 +11,5 @@ gem 'webrick', '~> 1.8'
 gem 'csv'
 gem 'base64'
 gem 'bigdecimal'
+
+gem "logger", "~> 1.7"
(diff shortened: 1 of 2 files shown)


adamscott (approved):

adamscott:
Thanks! 🚀

PR #1090, opened 2025-06-23
Fix 4.3 and 4.4 release pages

adamscott:
- Remove gsap
- Add anime.js (MIT licensed)


Calinou (approved):
Tested locally, it works as expected. This also makes files a good deal smaller.

For context, GSAP is under a [proprietary license](https://gsap.com/community/standard-license/). It also had issues when run locally but not on the production website (we couldn't find a solution). Switching to anime.js should resolve both issues at the same time.

File size before the PR (for GSAP + ScrollTrigger): 113.0 kB minified, 47.1 kB minified + gzipped
After the PR for anime.js: 76.5 kB minified, 27.3 kB minified + gzipped

PR #1092, opened 2025-06-23
Use "third-party" not "3rd party" in features.html

JohnVeness:
I think "third" looks more professional than "3rd". And "third-party" should be hyphenated in these particular instances because it's being used as an adjective.


coppolaemilio:
thank you :)

PR #1097, opened 2025-06-24
Make the check for `minify`/`gominify` binaries quiet

Calinou:
- Follow-up to https://github.com/godotengine/godot-website/pull/1096.

This prevents a message for `minify` not being found from being printed every time the website is built if you're using `gominify` instead. This still prints an error if neither `minify` or `gominify` is found.

This needs to be tested on Windows before merging. I assume it won't work, but it probably didn't work in `master` either as `which` is not a standard Windows utility (unless you happen to have Scoop installed and have run `scoop install which`).


akien-mga:
We should probably avoid merging PRs which break CI in the future ;)

coppolaemilio:
yeah, I didn't notice :(

PR #1103, opened 2025-06-25
Fix dotnet download button being too pale on the release pages

adamscott:
Currently, the dotnet download button on the release pages is too pale when hovered: it becomes pure white. This doesn't happen on the download page because the background is dark, and the gray of the button is actually transparent. As the background is white, it makes the button even bright.

The fix is to make the button darker.

| State | Before | After |
| :---: | :---: | :---: |
| Default | Capture d’écran, le 2025-06-25 à 11 53 32 | Capture d’écran, le 2025-06-25 à 11 53 21 |
| Hover | Capture d’écran, le 2025-06-25 à 11 40 46 | Capture d’écran, le 2025-06-25 à 11 52 45 |

### DIFF
--- a/assets/css/releases/4.3.scss
+++ b/assets/css/releases/4.3.scss
@@ -262,7 +262,7 @@ $donate-robot-size: 500px;
 			}
 
 			.download-net-button {
-				background-color: rgba(111, 111, 111, 27%);
+				background-color: rgba(0, 0, 0, 27%);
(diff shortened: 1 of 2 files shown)


adamscott:
@Calinou It seems that #1097 broke the test build.

Calinou (approved):
Code looks good to me.

PR #1137, opened 2025-08-11
Add gamescom community meetup to event list

winston-yallow:
Adds the community meetup to our events page.

The cover image is an edited version of the "Eingang Süd" image from the [gamescom image database](https://www.gamescom.global/en/info/press-and-creator/image-database) and credited according to their usage rules.

image

---

I have also included a small bugfix for missing whitespace between elements (notice the missing gap between the colon and the inline-paragraph that follows:

image

image


adamscott:
Added [a commit](https://github.com/godotengine/godot-website/pull/1137/commits/5997d6e03eda1e21147aed427737bf65ca3868ac) that reworks the events list.
- Add `cover_image_caption` (which uses [\<figcaption\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/figcaption) under the hood)
- Change event details from a list to a table (more semantic) (and looks better now)

Capture d’écran, le 2025-08-13 à 08 32 25

adamscott (approved):
Looks good to me!

PR #1139, opened 2025-08-14
Add the YouTube video to the "What's new in XR & Android" blog post

dsnopek:
The video of the presentation is now available, so this PR adds it to the article, and removes the note about how we'll add the video later


m4gr3d (changes requested):
There's a section at the end that needs to be updated as well.

dsnopek:
Ah, thanks, I've updated the bit at the end as well in my latest push

m4gr3d (approved):

PR #1146, opened 2025-08-23
showreels in lightbox

midzer:
Hi again,

Instead of opening YT showreels in a new tab, we could open them in the (already used) Tobii lightbox.

All functionality included (fullscreen, seek, volume etc). They're grouped and you can switch between them via navigation buttons.

Miserably, the navigation between the group is not working with keyboard arrow keys, because the YT object captures them. Swiping on mobile is not supported either. If you have an idea on how to fix this, please file in https://github.com/midzer/tobii upstream :)

Cheers
midzer


midzer:
Update tobii lightbox to v3 https://github.com/midzer/tobii/releases/tag/v3.0.0

* lightbox videos autoplay now on click
* `zoom: false` is default
* drop IE11 support
* lil smaller bundles

coppolaemilio:
I reviewed this one but I'm having a couple of issues in Firefox:
- When opening the videos, they don't show as big as they could
image
- Now it is hard to reproduce, but by playing a few of the videos and then closing the dialog, I could still hear the video playing even if it was not visible. 

I think the first one should be fixable, but the second one I will report more if I am able to reproduce it again.

midzer:
I've solved the first one.

Miserably, can't reproduce second one right now.

PR #1148, opened 2025-08-25
Added Godot Lyon's discord server

ysilog:
Adding my local godot community to the list.
This is a community based in Lyon (France) where we exchange about godot projects and meet monthly!


coppolaemilio:
Bienvenue!

PR #1157, opened 2025-09-05
Portuguese Localization

tiagosomda:
**disclaimer:** apologies if I forgot some specific step before creating this PR. This is my first time creating a PR for an open source project. I tried looking up what is Godot's specific guidelines and I think I was supposed to create an issue before creating the PR, but I felt like this was more for things like new features or modifying existing features.

## Summary
  - Added  Portuguese translation (_i18n/pt.yml)
  - Translated all home page content and included 'Pt' as a translated language.
  - Maintained technical accuracy while adapting content for Portuguese-speaking
  
## Translation Notes
Some English terms were intentionally preserved where the Portuguese versions are rarely used in the tech community, such as:
  - "blog" (commonly used as-is in Portuguese)
  - "web", "desktop", "mobile" (standard tech terminology)
  - "Assets" in "Biblioteca de Assets"


rsubtil:
Hi! I've read the proposed translation, and can confirm everything looks good! (I reviewed only the text itself; any technical requirements, I cannot review)

I've noticed that this seems to be Brazilian Portuguese, right? If yes, I'd say this should be internally renamed to `pt_BR` instead, similar to what's in the engine:
image

tiagosomda:
thanks for reviewing! good call on making sure it is similar to the engine.

speaking of 'internally' tho ... the url does end up being `pt_BR` as in `https://godotengine.org/pt_BR/` (which I consider it to be external facing xD)

I also noticed that other languages like French and Spanish don't have a country specific translation.
- It feels a bit odd to me seeing `pt_BR` ... makes me feel like we are excluding other Portuguese speaking groups 😅
- but it does keep it similar to the engine. 🤔

coppolaemilio:
Thank you so much for the contribution! So happy to finally have the time to review it. 
If we want to specify Brazil's locale, I think the standard to use here would be `pt-br` instead of `pt_BR`, but without wanting to pick up any fight, how different are they? We have only one version of `en` (even if there could be `en-us` and `en-uk`, etc...) one version of Spanish (same as en English), so I wonder, what would be a big difference here if we just keep one `pt` locale. Happy to merge/deploy this one soon :)

Calinou:
> but without wanting to pick up any fight, how different are they?

From what I know, they are pretty different in terms of idioms and all that. The difference is much greater than say, fr_FR versus fr_BE or even fr_FR versus fr_CA.

In terms of speaker count, pt_BR is much more prevalent than pt_PT, so it's probably fine to only have pt_BR at first.

tiagosomda:
thank you for the comments and review!
I changed to the standard format `pt-br`.
 
It seems that pt-br seems to be the most favorable right now.
I guess that once we end up having `pt-pt` (or others), we can change it up to support a generic one `pt` and regional ones too.

tiagosomda:
as for merging, I think we can squash merge it?
- no need to have these 5 commits as part of main's history

coppolaemilio:
obrigado!

tiagosomda:
de nada! 😁

PR #1166, opened 2025-09-16
Fix FR i18n entry for "header.features"

adamscott:
Replace for the FR i18n the term "Fonctions" with "Fonctionnalités" (for "header.features")

"Fonctions" means "Functions" instead of "Features" in this context. Like, it means more "what it does" as a manual tool than as "what it does" as a package. See [la Vitrine linguistique's entry](https://vitrinelinguistique.oqlf.gouv.qc.ca/fiche-gdt/fiche/26559312/fonctionnalite) on "functionnalité" as an equivalent term for "feature".

It was flagged to me by @gtibo.


Calinou (approved):
Looks good to me.

PR #1167, opened 2025-09-16
Fix header issues by relying on `isTouchDevice`

### LINKED ISSUE #1131: "Features" in the Header Menu opens the dropdown on click instead of on hover.
Currently the "Features" in the header will use the touch variant of it's handling on most devices.

The issue is caused by "isTouchDevice"

https://github.com/godotengine/godot-website/blob/627c5bee1e7875549414b6faa9f0757c2b8c9337/_includes/header.html#L229C1-L231C2

"maxTouchPoints > 0" checks, if the current browser supports touch inputs, not if touch is the currently used input method.

adamscott:
This PR makes the header code stop relying on `isTouchDevice` in order to change the behavior of the dropdown menus. The behavior is now different based on if the `PointerEvent` is of `pointerType` `"touch"` or otherwise.

Fixes #1131

https://github.com/user-attachments/assets/c6b31a01-acb4-4913-a868-ad30cf1bc7c4

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -230,9 +230,6 @@
 <script>
 	// Open dropdown on hover (desktop) or tap (touch devices)
 	document.addEventListener('DOMContentLoaded', function() {
-		// Detect if device has touch capability
-		const isTouchDevice = 'ontouchstart' in window || navigator.maxTouchPoints > 0;
-		
 		// Find all dropdown triggers
 		const dropdownTriggers = document.querySelectorAll('[data-dropdown]');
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Tested locally on desktop (no touch screen) and mobile, it works as expected. Behavior on desktop is the same as before, but mobile works correctly now. (That said, on mobile, I can't get the display to be wide enough to show the desktop navbar, even in landscape on a S25 Ultra).

Code looks good to me.

PR #1173, opened 2025-09-20
style: wrap showcase flip arrows in container for better visibility

SimonCastano:
Added a `.showcase-flip-container` around the left and right arrows.
This provides padding, rounded background, and hover effect to improve
their contrast and usability.


Calinou (commented):
Tested locally, it looks broken on my end on both Firefox 143 and Chromium 140:

image

The arrows appear on the left and are overlapping. It *is* possible to click the arrow that points to the left, but it's very hard.

PR #1192, opened 2025-10-17
Update build instructions to clarify that ruby 3.2+ is the minimum supported version now

### LINKED ISSUE #1190: README has conflicting information about minimum Ruby version
Under "Pre-requisites", it says

> Ruby 3.1.2 with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme)

Under "Building the website locally" it says:

> 1. Install Ruby 3.2 or later

But also:

>   - Set up `rbenv` running the following:
    - `rbenv install 3.4.4`
    - `rbenv global 3.4.4`

So which version should be used?

I ran into this because I can't currently build the website for some reason, so I am going over the build instructions carefully to see if I have something mis-configured. 

I suspect that the actual requirement is to use version 3.4.4 and the other mentions are out of date. I am hoping that's not the case because my OS doesn't ship with a new enough version of rbenv to support 3.4.4. So if that is correct, then I will need to build rebenv from source just to be able to test things that I write for the website

clayjohn:
Fixes: https://github.com/godotengine/godot-website/issues/1190

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ When working on new features[^2][^3], keep in mind this website only supports _e
 
 #### Pre-requisites
 - On Local Machine:
-  - Ruby 3.1.2 with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme)
+  - Ruby 3.2+ with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme)
   - Jekyll - [Jekyll Website](https://jekyllrb.com/)
   - Minify - [Minify GitHub](https://github.com/tdewolff/minify)
 - On Docker:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,12 +81,13 @@ To build the website locally, follow these steps:
   - If you don't have access to Ruby 3.2 or later in your distribution's repositories, install a compatible version of Ruby using `rbenv`:
     - Install in Ubuntu:
       - `sudo apt install rbenv`
+      - `sudo apt install libyaml-dev`
```
clayjohn, line 84: This is required for rbenv to build ruby 3.2.5. I'm sure there are other dependencies though. So we should consider enumerating all of the dependencies somehow instead of relying on contributors just happening to have the required dependencies. For now, this is an improvement though

README.md
```diff
@@ -81,12 +81,13 @@ To build the website locally, follow these steps:
...
     - Install in Fedora:
       - Install `rbenv` with dnf
-      - Run `echo 'eval "$(rbenv init -)"' >> ~/.bashrc` to add the rbenv init to `.bashrc` (or `.bash_profile`)
   - Set up `rbenv` running the following:
-    - `rbenv install 3.4.4`
-    - `rbenv global 3.4.4`
+    - `rbenv install 3.2.5`
```
clayjohn, line 88: We should be listing the minimum here instead of the latest to avoid confusion

README.md
```diff
@@ -81,12 +81,13 @@ To build the website locally, follow these steps:
...
+    - `rbenv global 3.2.5`
+    - `echo 'eval "$(rbenv init -)"' >> ~/.bashrc` to add the rbenv init to `.bashrc` (or `.bash_profile`)
```
clayjohn, line 90: This step is needed also needed on Ubuntu. So I moved it below

README.md
Calinou: I've tried to use `rbenv install 3.2.5` in a fresh Ubuntu 22.04 container. This is what I needed to get it to successfully compile:

```
sudo apt install build-essential libffi-dev libreadline-dev libyaml-dev zlib1g-dev
```

Note that I used upstream rbenv and [ruby-build](https://github.com/rbenv/ruby-build#readme) (what allows `rbenv install` to work), as opposed to the one that comes in the Ubuntu packages.

With this done, I was able to install `minify`, then run `rbenv global 3.2.5 && bundle install && bundle exec jekyll build`.

Calinou (approved):

PR #1203, opened 2025-11-01
Remove 'teams' page.

Ivorforce:
> [!NOTE]
> It has not been decided yet whether this is the best way to handle the problems at hand.

The teams page is very outdated, and it suggests contacting individual people for specific area work. RocketChat is more more appropriate for contact, since the activity of maintainers ebbs and flows, and specific area contributors might know more about something specific anyway.

The footer link, as well as stray visitors for the old page, will be redirected to the [contributing docs](https://contributing.godotengine.org/en/latest/organization/areas.html), which contains a more up-to-date view of the teams.

(we could also just remove the footer link?)

This removes the 'teams' tab from 'more':

SCR-20251101-jrjs


adamscott (changes requested):
We shouldn't remove `pages/teams.html` the current way it's done, as it would break every link that other people may have in other contexts. (i.e. somebody's presentation or something like that.)

Currently, the `pages/teams.html` already has a `redirect_to` parameter that redirects the user to the right page. So we could even let old links in the blogpost the same. I suggest that we don't change the contents of those either.

Ivorforce:
> We shouldn't remove `pages/teams.html` the current way it's done, as it would break every link that other people may have in other contexts. (i.e. somebody's presentation or something like that.)

Shouldn't those be redirected to the new page automatically? What are you suggesting we do instead?

> Currently, the `pages/teams.html` already has a `redirect_to` parameter that redirects the user to the right page. So we could even let old links in the blogpost the same. I suggest that we don't change the contents of those either.

The reason I edited those is that, whenever the link of the new page might change (e.g. from `/organization/teams` to `/general/teams`, or anything else), it would be easier to find and replace those links again. And when we finally remove the redirect from the old site, there's not a bunch of links to backtrack internally.
But, I'm not set in stone on this. We can leave them as-is if you prefer.

adamscott:
> Shouldn't those be redirected to the new page automatically? What are you suggesting we do instead?

I mean how it will be redirected if we remove the page that it links to and we don't tell the static site to redirect people to the new page? I may miss something, but if we remove the page, it will just not exist anymore at all, even the redirection.

cc. @coppolaemilio

Ivorforce:
> I mean how it will be redirected if we remove the page that it links to and we don't tell the static site to redirect people to the new page? I may miss something, but if we remove the page, it will just not exist anymore at all, even the redirection.

I'm still not entirely sure I understand what you're saying.
The `pages/teams.html` page still exists in this PR. I'm just removing the content, not the page itself — And I'm adding a `redirect_to` to the contributing docs team page. People with links to the current page should be redirected automatically to the new page.

adamscott:
> > I mean how it will be redirected if we remove the page that it links to and we don't tell the static site to redirect people to the new page? I may miss something, but if we remove the page, it will just not exist anymore at all, even the redirection.
> 
> I'm still not entirely sure I understand what you're saying. The `pages/teams.html` page still exists in this PR. I'm just removing the content, not the page itself — And I'm adding a `redirect_to` to the contributing docs team page. People with links to the current page should be redirected automatically to the new page.

Sorry. I mistook the large diff with the removal of the file. Sorry.

Ivorforce:
No worries, I'm just happy I understand now 😄

coppolaemilio (approved):

coppolaemilio:
Thank you 🫡

PR #1207, opened 2025-11-06
Update featured images

brcontainer:
Think of this as a draft for now; it may be merged into the final version after the release of version 4.6.


coppolaemilio:
Thanks! I will merge now since we are preparing the site for the release :)

PR #1227, opened 2025-12-16
Convert every WebP cover to JPG for compatibility purposes

adamscott:
> [!NOTE]
> There was a lot of issues with cover images not displaying on certain websites, such as on Reddit.

😢

Also add a new filter, `ensure_magic_type` (`_plugins/ensure_magic_type.rb`), that checks the input magic type and compares it to its parameter. If it doesn't match, it fails the build. There's an optional context parameter to explain the check. 

Used in `_layouts/default.html` for the cover image metadata where it bans WebP files.


coppolaemilio (changes requested):
Hey! I like the idea of the `ensure_magic_type` warning, but we have already done a few "let's convert everything to X format" before and it is quite a rocky ride each time. Would it be possible to leave this PR to be just about future additions and not trying to convert every webp that already exists? Don't want to end up having deep fried images after a few more changes to webp and back from it to other format 😅

adamscott:
Superseded by #1229.

PR #1229, opened 2025-12-17
Ensure new cover images only use basic image formats

adamscott:
- Add new filter `ensure_magic_type` (`_plugins/ensure_magic_type.rb`) that checks the input magic type and compares it to its parameter. If it doesn't match, it fails the build. There's optional context parameters to explain the check.

- Add new filter `is_legacy_cover_image` (`_plugins/is_legacy.rb`) that is essentially a grand-fathered clause check to see if we apply a `ensure_magic_type` check for the cover image.

Supersedes #1227 by adding a legacy check and by not converting every existing cover image to JPG.


_plugins/ensure_magic_type.rb
```diff
@@ -0,0 +1,110 @@
...
+
+module Jekyll
+  module EnsureMagicTypeFilter
+    @@known_types = {
+      "jpeg" => "JPEG image data",
+      "jxl" => "JPEG XL codestream",
+      "png" => "PNG image data",
+      "gif" => "GIF image data",
+      "webp" => "Web/P image",
+    }
```
Calinou, line 25: Is it possible to add AVIF detection?

Calinou (approved):
Tested locally, it works as expected. Code looks good to me.

_plugins/ensure_magic_type.rb
adamscott: Done!

coppolaemilio (approved):

coppolaemilio:
Thank-you.jpg 🤓

PR #1238, opened 2026-01-09
Update Priorities page to match status as of January 2026

Calinou:
Priorities that were completed were removed[^1], while some other priorities were updated to match the current development status.

This is based on https://github.com/godotengine/godot-website/issues/1232, but also a previous effort where various core contributors (myself included) have combed through the current priorities and detailed their current status.

- This closes https://github.com/godotengine/godot-website/issues/1232.

[^1]: There is a `status` line, but there is no different styling for "done" items currently. This would have to be implemented first if we want to keep listing done items, although I'm not sure it's really needed.

PR #1241, opened 2026-01-17
Remove finished priorities in XR and GDExtension

dsnopek:
I was looking at the priorities page, and noticed that some of the things listed for XR and GDExtension are done! (Although, some of them will only be first released with Godot 4.6.)

I also saw that one of the linked GDExtension PRs was unrelated to the priority it was linked from and removed it.


Calinou:
Note that https://github.com/godotengine/godot-website/pull/1238 handles this as well. I forgot to remove some priorities that this PR removes, so I went and updated my PR accordingly. Thanks!

Note that I kept the LibGodot priority, but reworded it to be about improving the existing LibGodot support.

> I also saw that one of the linked GDExtension PRs was unrelated to the priority it was linked from and removed it.

I just did this as well in my PR.

dsnopek:
Ah, thanks, I hadn't noticed that PR!

Superseded by https://github.com/godotengine/godot-website/pull/1238

PR #1256, opened 2026-02-05
Split releases on home page news section

coppolaemilio:
Since we publish way more pre-releases than anything else, it is very common for them to be hiding our regular blog posts. This change splits the home page section into news + releases, and pre-releases + releases:

image

This is just the first step in many changes we are experimenting to refresh how we present new snapshots. 

I had to create two localization keys so if you speak any of the languages we currently support please make sure to review those translations as well, they are currently automatic.


Calinou (approved):
Looks good to me. French translation is correct as well.

PR #1265, opened 2026-02-18
Added the div.add-nav-gap to user-groups.html (avoid overlap top menu)

mdtrooper:
Add spacing offset (.add-nav-gap) in user-groups.html to avoid header overlapping content


coppolaemilio:
gracias!

PR #1068, opened 2025-05-16
Persian translation

AturAseman:
not sure if there's RTL support yet but I'm leaving this here for when/if it's added. also, a standard web font you could use for Arabic script and its varieties is Vazirmatn (https://rastikerdar.github.io/vazirmatn/en).

PR #1277, opened 2026-03-03
Add persian translation (rebased)

Ivorforce:
Supersedes #1068, created by @AturAseman. I didn't check translations for correctness.

I merely squashed and resolved conflicts, but wasn't able to push to their branch. Commit is attributed to them.


Ivorforce:
Public update: This appears to be blocked because we practically can't really add more languages right now. This is up to the website / infrastructure team to fix before we can merge this.

PR #1286, opened 2026-03-11
Fix some misc issues

adamscott:
- Removes fake paragraph separators (`<br><br>`) in the Features page to replace them with real semantic paragraphs.
- Adds `features.scss` while at it (with a new sass module, `common/utils`)
- Adds `rdoc` to bundle as it was removed in Ruby 4.0 (it was built-in before)


coppolaemilio (approved):
I wanted us to move away from scss eventually, but the changes here look good

Calinou (approved):
Tested locally, it works as expected. Code looks good to me.

PR #1294, opened 2026-03-21
remove itch.io link for KinitoPET

reckdave:
The only official download for KinitoPET is on Steam page where you can obtain the game legally. The itch.io version may contain malware which is a huge risk if someone downloads this as its directly from the godots official website,, if it doesnt have malware then its still considered piracy either way. Troy's website/portfolio [https://l-l.dev/] does NOT include this version so im sure that this isnt official.


akien-mga (approved):

akien-mga:
Thanks!

PR #1301, opened 2026-03-30
Add Godoteires (Galician Godot User Group)

NEKERAFA:
Fixed also a bug in Zenvoid coordinates that caused an error in JavaScript and prevented online communities from displaying.


coppolaemilio:
grazas! :)

PR #1309, opened 2026-04-10
Unify `versions.yml` syntax (reverted)

Repiteo:
In order to prevent further regressions in `versions.yml`, this aims to eliminate any remaining discrepancies by having ALL releases consolidated to the array. If something needs to pick from specifically the latest iteration, it can do so by accessing the first array element. Documentation has been updated accordingly.


akien-mga:
Can you confirm that the generated `versions.json` is identical to https://godotengine.org/versions.json ?

Repiteo:
It's not 100% identical, but that's only because it no longer excludes the `featured` or `release_version` variables. Those being excluded were probably an oversight to begin with, so I'd say this was an improvement

akien-mga:
> It's not 100% identical, but that's only because it no longer excludes the `featured` or `release_version` variables. Those being excluded were probably an oversight to begin with, so I'd say this was an improvement

@KoBeWi Do you foresee any issue for the version checker or it's fine if there are additional fields in the json?

KoBeWi:
Extra fields won't be a problem.

akien-mga (approved):

akien-mga:
Thanks!

akien-mga:
Seems like this broke the download link generator:
https://www.reddit.com/r/godot/comments/1sivx1c/godot_latest_net_version_not_downloading_from_the/

```
_plugins/make_download.rb
109:      return "#{HOST_HETZNER}?version=#{version_name}&flavor=#{version_flavor}&slug=#{download_slug}&platform=#{platform}"
```

There's still a bunch more occurrences of `flavor` in the codebase which need to be handled.

For example for the /download/preview page: https://godotengine.org/download/preview/

I'll revert for now to fix downloads.

Repiteo:
> There's still a bunch more occurrences of `flavor` in the codebase which need to be handled.

Frustratingly, that doesn't necessarily correlate to code that needs to be changed; `flavor` is what the struct uses, so a TON of false-positives crop up. I'll give this another pass in those areas though

PR #1314, opened 2026-04-23
Fix Windows PR link in HDR output article

ProgrammerOnCoffee:
Looks like the HDR output article links to the Linux PR twice.


Calinou (approved):
Looks good to me.

PR #1317, opened 2026-04-24
Update 4.7 beta 1 with volume calculation fix

Ivorforce:
Ref: https://github.com/godotengine/godot/pull/114080

This is fairly major for any games that need multiple viewports. I believe it's a straight-up upgrade, but it _will_ be a major difference for any games that are affected.

cc @Repiteo


Repiteo:
Thanks!

PR #1319, opened 2026-04-29
Fix date on HDR output blog post and add Stuart as a contributor.

allenwp:
### What problem(s) does this PR solve?

1. Date is incorrect on HDR output blog post (it went live sometime around noon Eastern Daylight Time on April 23)
2. Of the major platform PR authors, Stuart is the only contributor who is not highlighted

### Additional information

With so many people helping with the development of the HDR output feature, it's hard to figure out where the line should be regarding which contributors are highlighted in the blog post. But I think it's reasonable to highlight Stuart for his contributions to adding HDR output support for macOS, iOS, and visionOS.


Repiteo (approved):

Repiteo:
Thanks!

PR #1345, opened 2026-06-01
Add GABE to Android download page

Calinou:
- This closes https://github.com/godotengine/godot-website/issues/1339.

## Preview

It's displayed in the *Other Godot downloads* section, but only when the Android download page is being viewed (as it's not relevant for other platforms):

image

It could be moved to the main section with the editor, but the data source isn't designed to accomodate something with a different versioning scheme. Also, from an UI perspective, it could lead to some people downloading GABE instead of the editor.

PR #1346, opened 2026-06-01
Update Gemfile for Ruby 4

Calinou:
This is required for `bundle install` to successfully build native extensions on Linux (it failed with `google-protobuf` 4.31.4).

Tested on a fresh Ubuntu 24.04 Docker container with Ruby 3.2.3, as well as a Fedora 43 system with Ruby 4.0.1.

I needed this while developing #1345.


adamscott (approved):
Thanks! I think it's the right thing to do.

PR #1375, opened 2026-06-24
Fix (for good) the comparison slider mobile issues

adamscott:
Follow-up to #1372

This definitely fixes the release page comparison slider issues as it makes sure to prevent the default behavior of the touch events emitted.

Now, when dragging the slider, the scrolling is completely disabled.

### DIFF
--- a/assets/css/releases/components/media_block.css
+++ b/assets/css/releases/components/media_block.css
@@ -73,6 +73,10 @@
     width: 100%;
     height: 100%;
     touch-action: pan-y pinch-zoom;
+	* {
+		pointer-events: none;
+		user-select: none;
+	}
     img{
       user-select: none;
       pointer-events: none;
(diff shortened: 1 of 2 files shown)


assets/css/releases/components/media_block.css
```diff
@@ -73,6 +73,10 @@
     width: 100%;
     height: 100%;
     touch-action: pan-y pinch-zoom;
+	* {
```
coppolaemilio, line 76: isn't using `*` a bit risky here? could any of these properties affect other elements we already have on the website?

coppolaemilio (approved):

assets/css/releases/components/media_block.css
adamscott: In this context, it matches only every element inside `.media-compare` (`.media-block .media-compare *`), not every elements of the website (fortunately).

PR #1380, opened 2026-06-24
Update README to use Ruby 4.0

adamscott:


README.md
```diff
@@ -58,7 +58,7 @@ When working on new features[^2][^3], keep in mind this website only supports _e
 
 #### Pre-requisites
 - On Local Machine:
-  - Ruby 3.2+ with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme)
+  - Ruby 4.0+ with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme)
```
Calinou, line 61:
```suggestion
  - Ruby 3.2+ with [rbenv](https://github.com/rbenv/rbenv) and [*(For windows)*](https://github.com/RubyMetric/rbenv-for-windows#readme) (4.0+ recommended)
```

Calinou (commented):
As per https://github.com/godotengine/godot-website/pull/1346, there is no strict requirement on Ruby 4.0 yet. That said, I agree it's better to recommend Ruby 4.0 to be ready for future upgrades (as I expect we'll require it eventually).

PR #1407, opened 2026-07-29
Fix reflows when hovering over headings due to blog anchor links

### LINKED ISSUE #1406: When the section link appears, it jumps
This is a fairly simple but bad-for-readability bug. When hovering over a link, the styles are set up in such a way that the block expands and abruptly pushes the blocks below it down.

I noticed this by accident. It interferes with reading because, while reading, you naturally hover over it unintentionally, and the page starts jumping: first downward, and then, when the mouse moves away, it shifts back upward. This is an automatic action that happens during reading, and it disrupts the reading experience.

It can be fixed in different ways. I don't know which solution would work best for you.

https://github.com/user-attachments/assets/52da027b-072e-4826-99a3-394d8175f131

The place where it can be seen: https://godotengine.org/article/dev-snapshot-godot-4-8-dev-2/

Calinou:
This uses CSS `visibility` instead of `display` to toggle whether the anchor link is visible, so that it always takes up space in the layout.

- This closes https://github.com/godotengine/godot-website/issues/1406.

### `master`

https://github.com/user-attachments/assets/d2dc010f-a44b-4507-b9be-c1141e8498bd

### This PR

https://github.com/user-attachments/assets/361b1418-e359-4446-8577-5597a06b843a

### DIFF
--- a/assets/css/anchor-link.css
+++ b/assets/css/anchor-link.css
@@ -1,5 +1,6 @@
 .anchored-link {
-    display: none;
+    display: inline-block;
+    visibility: hidden;
     background-image: url(/assets/icons/link.svg);
     background-repeat: no-repeat;
     background-position: center;
@@ -26,7 +27,8 @@
 }
 
 .anchored:hover .anchored-link {
-    display: inline-block;
+    /* Change `visibility` instead of `display` to avoid reflows when hovering over the link. */
+    visibility: visible;
 }
 
 .anchored-toast {


JohnVeness:
I like the idea of this, but it does seem a shame that this introduces blank vertical space underneath headings, where affected. Would it be possible to put the icons in the margin to the left of the heading instead?

Calinou:
> I like the idea of this, but it does seem a shame that this introduces blank vertical space underneath headings, where affected. Would it be possible to put the icons in the margin to the left of the heading instead?

That was my other idea, but I'm concerned it won't look great on mobile or on narrow displays where there's little margin on the left:

image

JohnVeness:
I suppose alternatively the icon could overlap the heading text if it wouldn't otherwise fit to the right of it (obviously, only when hovered over), e.g.:

godot

(although prettier, maybe with a nice fade effect).

PR #1417, opened 2026-08-14
Add showreel submission opening for 2026

coppolaemilio:
To be published tomorrow (Aug 15, 2026).

Image might be also updated depending if I can get a new one in time. 

The submissions are already open, but there isn't any issue about that, technically people can already submit entries.

### DIFF
--- a/pages/community/community.html
+++ b/pages/community/community.html
@@ -159,8 +159,8 @@ layout: default
 				
 				<div class="events-upcoming">
 					<span>Next event:</span>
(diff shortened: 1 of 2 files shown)


pages/community/community.html
JohnVeness, line 1: This file shouldn't really by part of this PR, should it?

coppolaemilio: nah, but I thought I might as well fix that one while I was at it :)

